Issue #8, opened 2022-03-30
Test result not found error when running any test


georgemickleburgh:
Hi, I'm trying to get the extension to run on Windows and I seem to be getting the same "Test result not found" error when running any of the tests.

I did a debug run of a test to see what was running and the output was:

```
"E:\Program Files\nodejs\node.exe" .\node_modules\vitest\vitest.mjs run /Users/George/frontend-v8/src/composables/__tests__/auth.test.ts --testNamePattern "auth composable should start with a null current user and account" --reporter=default --reporter=json --outputFile C:\Users\George\AppData\Local\Temp\vitest-report-b18l2.json

 RUN  v0.8.0 c:/Users/George/frontend-v8

Could not read source map for file:///c:/Users/George/frontend-v8/@vite/env: ENOENT: no such file or directory, open 'c:\Users\George\frontend-v8\@vite\env.mjs.map'
⎯⎯⎯⎯⎯⎯ Failed Suites 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/composables/__tests__/auth.test.ts [ src/composables/__tests__/auth.test.ts ]

Error: No test suite found in file c:/Users/George/frontend-v8/src/composables/__tests__/auth.test.test

 ❯ runFiles node_modules/vitest/dist/entry.js:870:18
 ❯ startTests node_modules/vitest/dist/entry.js:880:9
 ❯ async node_modules/vitest/dist/entry.js:907:7
 ❯ async withEnv node_modules/vitest/dist/entry.js:504:5
 ❯ async run node_modules/vitest/dist/entry.js:906:5
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

Test Files  1 failed (1)
     Tests  no tests
      Time  5.98s (in thread 0ms, Infinity%)
JSON report written to C:/Users/George/AppData/Local/Temp/vitest-report-b18l2.json
Process exited with code 1
```

If I run the exact same command within from the Terminal inside VSCode I get the expected result:

```
[...]

PR #10, opened 2022-03-30
fix: force spawn to use powershell on windows and fixed drive case match

georgemickleburgh:
This PR fixes the issue that I had in #8 on my Windows machine. 

The first part is forcing the shell to be `powershell` rather than the default value of `true`, which I'm guessing uses `cmd` (I couldn't find any good docs on that). I don't think this should cause any issues on other Windows machines, but I'm unable to test.

The second part is that the tests report their results in the file as `X:/path_to_test`, but the `vscode.TestItem[]` seems to input the files with a lowercase `x:/path_to_test`, so I correct it in the Windows part of the `pathToFile` map to allow the tests to be correctly identified.

### DIFF
--- a/src/pure/runner.ts
+++ b/src/pure/runner.ts
@@ -101,7 +101,7 @@ export class TestRunner {
         cwd: workspacePath,
         stdio: ["ignore", "pipe", "pipe"],
         env,
-        shell: isWindows,
+        shell: isWindows ? 'powershell' : false,
       });
 
       for await (const line of chunksToLinesAsync(child.stdout)) {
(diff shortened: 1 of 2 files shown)


zxch3n:
Thanks! @georgemickleburgh  LGTM

xon52:
Investigating issues I have found that forcing the shell to be `powershell` is completely failing the command and no report is generated. Only `true` works for me on Windows.
It may be that I use Bash as my default? I wonder if `true` is just use your default and vscode has an issue? I know when they updated the shell options list they broke something, but then fixed it. Maybe it wasn't completely fixed.

Either way, I think this change needs more testing.

Issue #15, opened 2022-04-11
vitest not work - "( Vitest should be configured to be able to run from project root )"


mengqianyixing:
![image](https://user-images.githubusercontent.com/70135517/162660924-7a9d3269-622c-4363-9694-d322183ec9af.png)

Aaron-Pool:
@mengqianyixing 

I got this when I first tried to use the extension as well. The solution for me was to use the following setting to have my package manager run the test instead of whatever the extension was trying to do to locate the correct command:

![image](https://user-images.githubusercontent.com/5634834/163435316-5992b009-a163-43ab-9ae1-1248af5f5a5e.png)

In my case, `pnpm vitest` was the correct setting. But yours may be different, depending on the package manager you're using and what it's cli is for using installed binaries.

xon52:
`pnpm` didn't help; I'm still having this error.

mengqianyixing:
> @mengqianyixing
> 
> I got this when I first tried to use the extension as well. The solution for me was to use the following setting to have my package manager run the test instead of whatever the extension was trying to do to locate the correct command:
> 
> ![image](https://user-images.githubusercontent.com/5634834/163435316-5992b009-a163-43ab-9ae1-1248af5f5a5e.png)
> 
> In my case, `pnpm vitest` was the correct setting. But yours may be different, depending on the package manager you're using and what it's cli is for using installed binaries.

it's still not work by use 'pnpm vitest', look like the "vitest" it's not good support for the windows. :)

Aaron-Pool:
> > @mengqianyixing
> > I got this when I first tried to use the extension as well. The solution for me was to use the following setting to have my package manager run the test instead of whatever the extension was trying to do to locate the correct command:
> > ![image](https://user-images.githubusercontent.com/5634834/163435316-5992b009-a163-43ab-9ae1-1248af5f5a5e.png)
> > In my case, `pnpm vitest` was the correct setting. But yours may be different, depending on the package manager you're using and what it's cli is for using installed binaries.
> 
> it's still not work by use 'pnpm vitest', look like the "vitest" it's not good support for the windows. :)

For what it's worth, I'm on windows as well. And it's working for me. So I'm guessing windows isn't the root of your problem.

zxch3n:
@mengqianyixing Is this still happening on v0.1.17?

AlmarAubel:
We still have the problem in v0.1.17

zxch3n:
[...]

PR #18, opened 2022-04-18
fix: use 'node vitest.mjs' as default vitest command on windows

InfiniteXyy:
This PR fixes a small bug on Windows (with default settings).

Without any settings, the vitestCommand in `TestRunner`  class will be a `.mjs` file, which is not callable on windows.  So I change it to `node vitest.mjs` on windows by default, and on other operation system, the ./vitest.mjs command should work. 

I'm not quite sure, if this will fix the problem mentioned in #15, but it do solve my problem.


zxch3n:
Thanks!

PR #23, opened 2022-04-23
fix debugging on windows

AlmarAubel:
Because debugger is running within vscode without using a the powershell shell we need to bypass the drive letter fix while debugging.

This will fix the issue mentiond in #15

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "vitest-explorer",
   "displayName": "Vitest",
   "description": "Run and debug Vitest test cases",
-  "version": "0.1.18",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "vitest-explorer",
   "displayName": "Vitest",
   "description": "Run and debug Vitest test cases",
-  "version": "0.1.18",
+  "version": "0.1.19",
```
zxch3n, line 5: This change is not necessary

zxch3n:
Great. It works

PR #24, opened 2022-04-24
File path corrections for Windows

xon52:
I have not found that the latest updates have helped, but I did find that the path issues identified by @AlmarAubel are the issue. These changes make a single path sanitization function which has solved the issues for `runTest()` (normal) on my Windows 10 machine.

#### I have still been unable to run/fix the `debugTest()` (debug) for my windows machine

### DIFF
--- a/src/pure/runner.ts
+++ b/src/pure/runner.ts
@@ -13,7 +13,9 @@ export function getDebuggerConfig() {}
 let i = 0;
 const suffix = (0 | (Math.random() * 1000000)).toString(36);
 export function getTempPath(): string {
-  return sanitizeFilePath(path.join(tmpdir(), `vitest-report-${suffix}${i++}.json`));
+  return sanitizeFilePath(
+    path.join(tmpdir(), `vitest-report-${suffix}${i++}.json`),
+  );
 }
 
 type Status = "passed" | "failed" | "skipped" | "pending" | "todo" | "disabled";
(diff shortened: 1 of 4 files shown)


AlmarAubel:
It looks almost the same as me first try to fix this issue (see this [commit](https://github.com/vitest-dev/vscode/commit/b484b453d0e3ccff028544dbcd1fa0d06306810f)), although you made some nice additional refactoring ;). But this was not working for every one. 

Altough your fix is working on my windows computer. But I'm not sure if some functions will be broken if you don't pass  the env vars to the spawn command.

xon52:
@AlmarAubel I'll take it out the changes to `spawn()` when I update with the debug updates; I didn't mean to leave them in. 👍 

Might make another PR for it though as I cannot see what they do. Not sure where it comes from or why it's needed. Seems to work on Mac and Windows and may be leftover boiler plate or a mask for a solved issue 🤷‍♂️

src/pure/runner.ts
```diff
@@ -93,17 +87,15 @@ export class TestRunner {
...
     let error: any;
     let outputs: string[] = [];
     const env = { ...process.env, ...workspaceEnv };
     try {
       // it will throw when test failed or the testing is failed to run
       const child = spawn(command, args, {
         cwd: workspacePath,
-        stdio: ["ignore", "pipe", "pipe"],
-        env,
-        shell: isWindows ? "powershell" : false,
```
xon52, line 106: @zxch3n when I remove these and leave shell as `true` it seems to work well. Do you know where these args come from and why they're there?

AlmarAubel: I think they are used to pass true some custom settings from the settings.json

image

zxch3n: Why remove `stdio` and `env`? stdio is used to get the test output. env for nodeEnv config

xon52: No problem. Put them back 😀

xon52:
@zxch3n looks good and works fine for a normal run. Obviously debug still does not work on my machine but the primary function works well.
I made one small commit to make a string nicer but it should not affect anything I think.

PR #39, opened 2022-05-23
Feature suggestion. Add failure notification using vscode message. Optional using extension setting.

Tocseoj:
I like the Watch Mode for Vitest and thought it would be a nice feature if when making changes you got an immediate pop-up on a test failure, so you know the change that broke it. I made the pop-up an optional  boolean setting.

Screen Shot 2022-05-23 at 3 01 55 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,12 @@
           ],
           "default": null,
           "scope": "machine"
+        },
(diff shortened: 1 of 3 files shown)


zxch3n:
Thanks!

PR #40, opened 2022-05-24
feat: multi-root workspace support

evanandrewrose:
Hey,

Thanks for the extension. It's been great to have.

I recently started using [multi-root workspaces](https://code.visualstudio.com/docs/editor/multi-root-workspaces) to manage by related repos and I noticed that this plugin didn't support that use case.

This PR adds support by:

In the case of **running tests**, start a `TestRunner` for each workspace. This allows you to shift-click all tests across multiple workspaces and run them as though they're in the same workspace.
In the case of **watching tests**, I just mapped over a vitest configuration derived from each workspace root, meaning that multiple vitest instances will be spawned (1 per vitest-supported root).
In the case of **debugging tests**, the code assumes you're not debugging across multiple workspace roots and it chooses the root associated with the first test in the list when figuring out which vitest instance to invoke.


zxch3n:
Great job, thanks!

src/extension.ts
```diff
@@ -48,46 +48,55 @@ export async function activate(context: vscode.ExtensionContext) {
...
-  const vitestVersion = await getVitestVersion(vitestCmd, getConfig().env || undefined).catch(async (e) => {
-    log.appendLine(e.toString())
-    log.appendLine(`process.env.PATH = ${process.env.PATH}`)
-    log.appendLine(`vitest.nodeEnv = ${JSON.stringify(getConfig().env)}`)
-    let errorMsg = e.toString()
-    if (!isNodeAvailable(getConfig().env || undefined)) {
-      log.appendLine('Cannot spawn node process')
-      errorMsg += 'Cannot spawn node process. Please try setting vitest.nodeEnv as {"PATH": "/path/to/node"} in your settings.'
-    }
+    const version = await getVitestVersion(cmd, getConfig().env || undefined).catch(async (e) => {
```
zxch3n, line 59: `getConfig` should have a WorkspaceFolder param now, to use it in the workspace.getConfiguration

```ts
export function getConfiguration(section?: string, scope?: ConfigurationScope | null): WorkspaceConfiguration;
```

`!getConfig().enable` above need to be changed too

src/extension.ts
```diff
@@ -48,46 +48,55 @@ export async function activate(context: vscode.ExtensionContext) {
...
-    const msg = 'Because Vitest version < 0.8.0, run/debug/watch tests from Vitest extension disabled.\n'
+
+  const isCompatibleVitestConfig = (config: typeof vitestConfigs[number]) =>
+    (config.version && semver.gte(config.version, '0.8.0')) || customTestCmd
+  const isNotCompatibleVitestConfig = (config: typeof vitestConfigs[number]) =>
+    !isCompatibleVitestConfig(config)
+
+  vitestConfigs.filter(isNotCompatibleVitestConfig).forEach((config) => {
+    const msg = `Because Vitest version < 0.8.0 in ${config.workspace.name}, run/debug/watch tests from Vitest extension disabled.\n`
     context.subscriptions.push(
```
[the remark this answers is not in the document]
evanandrewrose: Good point, fixed.

package.json
```diff
@@ -119,6 +119,12 @@
           ],
           "default": null,
           "scope": "resource"
+        },
+        "vitest.showFailMessages": {
+          "description": "Get instant feedback when using Watch Mode. Pop-ups an error when a test fails.",
+          "type": "boolean",
+          "scope": "window",
```
evanandrewrose, line 126: note: changing this to window to prevent users from incorrectly trying to set this value at the workspace folder level

src/config.ts
```diff
@@ -29,3 +29,11 @@ export function getConfig(workspaceFolder: WorkspaceFolder) {
     enable: get<boolean>('enable', false),
   }
 }
+
+export function getRootConfig() {
```
evanandrewrose, line 33: adding this to differentiate from potentially folder-level configs

src/runHandler.ts
```diff
@@ -63,7 +75,7 @@ export async function updateSnapshot(
 
   test = testItemIdMap.get(ctrl)!.get(test.id)!
   const runner = new TestRunner(
-    vscode.workspace.workspaceFolders[0].uri.fsPath,
+    determineWorkspaceForTestItems([test], vscode.workspace.workspaceFolders).uri.fsPath,
     getVitestCommand(vscode.workspace.workspaceFolders[0].uri.fsPath),
```
zxch3n, line 79: should use corresponding workspace folder here

src/config.ts
```diff
@@ -1,14 +1,39 @@
...
 
-export function getConfig() {
-  const config = workspace.getConfiguration('vitest')
+export function getConfigValue<T>(
+  rootConfig: WorkspaceConfiguration,
+  folderConfig: WorkspaceConfiguration,
+  key: string,
+  defaultValue: T,
+): T {
+  return folderConfig.has(key)
```
zxch3n, line 11: Can this be simplified as `folderConfig.get(key) ?? rootConfig.get(key) ?? defaultValue`

src/extension.ts
```diff
@@ -19,10 +19,10 @@ export async function activate(context: vscode.ExtensionContext) {
   )
     return
 
-  if (
-    !getConfig().enable
-    && !(await isVitestEnv(vscode.workspace.workspaceFolders[0].uri.fsPath))
-  )
+  const vitestEnvironmentFolders = vscode.workspace.workspaceFolders.filter(async folder =>
```
zxch3n, line 22: filter cannot take async function as param. async func will always return Promise, which will be taken as True in filter

src/runHandler.ts
```diff
@@ -91,12 +103,41 @@ export async function debugHandler(
   run.end()
```
zxch3n, line 103: debugHandler need the same pre-process in runHandler. Because `runTest` assumes all the tests come from the same workspace folder

evanandrewrose: yeah, that's a good idea

src/watch.ts
```diff
@@ -70,10 +74,10 @@ export class TestWatcher extends Disposable {
       let timer: any
       this.process = execWithLog(
         this.vitest.cmd,
-        [...this.vitest.args, '--api', port.toString()],
+        [...this.vitest.args, '--api.port', port.toString()],
```
zxch3n, line 77: why append `.port`

src/runHandler.ts
```diff
@@ -35,19 +35,29 @@ export async function runHandler(
...
-  const tests = request.include ?? gatherTestItems(ctrl.items)
-  const runner = new TestRunner(
-    vscode.workspace.workspaceFolders[0].uri.fsPath,
-    getVitestCommand(vscode.workspace.workspaceFolders[0].uri.fsPath),
-  )
-
   const run = ctrl.createTestRun(request)
-  await runTest(ctrl, runner, run, tests, 'run')
+
+  for (const folder of vscode.workspace.workspaceFolders) {
```
zxch3n, line 48: The previous `await Promise.all` for runHandler did make sense to me. Because the running order doesn't matter when running tests in different workspaces, though it does when debugging

The previous suggested change was for `debugHandler`.

Issue #52, opened 2022-07-08
Debugging cannot find any test files


thepaperpilot:
**Describe the bug**
When telling VSC to debug the tests (all of them, a single test, a single file, or a single describe), will result in nothing happening, apart from a log in the debug console saying no test files were found.

**To Reproduce**
1. Go to the testing panel
2. Click the "Debug Tests" button
3. If its not open already, open the debug console

**Expected behavior**
The tests to run, using the debugger to stop at breakpoints and such

**Screenshots**
```bash
C:\Program Files\nodejs\node.exe .\node_modules\vitest\vitest.mjs vitest D:/projects/Kronos/tests/util/common.test.ts --api 65426
 DEV  v0.17.1 d:/projects/Kronos
      API started at http://localhost:65426
filter:  vitest, D:/projects/Kronos/tests/util/common.test.ts
include: **/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}
watch exclude:  **/node_modules/**, **/dist/**

No test files found, exiting with code 1
Process exited with code 1
```

**Environment**

- OS: windows
- VSCode version: 1.69.0
- Vitest version: 0.17.1
- Vitest plugin version: 0.2.20

**Additional context**

Using the launch.json config to run tests on the open tests file works as expected. I noticed there appears to be an extra `vitest` in the command its running in the debug console, which then appears in the filter list. I suspect this is the root of the issue - it's filtering for test files that include "vitest" in their name, which is none of them.

BePo65:
I have the same problem; can you help out with the contents of the `launch.json` you mentioned above?

thepaperpilot:
Sure thing: it's the one in the docs [here](https://vitest.dev/guide/debugging.html).

devatina11yb:
Any progress on this issue?

There is actually a workaround which would be to add `vitest ` to the name of every test file (e.g. `vitest file.spec.ts`) but there should be a cleaner fix. Also, I am unsure if it is related but watch mode doesn’t seem to be working.

countnazgul:
Haven't investigated why but downgrading `vitest` itself to `0.31.0` fixed it for me. For reference: the latest version of `vitest` at the moment is `0.31.1`

I was getting the same issue `No test files found` but in my case the `filter` property was not including the drive path:
My tests were in `D:/my-project/test` and the `filter` property was showing `/my-project/test` (not sure if this was the real issue or it was just display issue)

[...]

PR #56, opened 2022-07-14
handle the case where the command starts with a lower case driver letter

lbm98:
When running tests on my machine, I got errors saying:

> Test result not found.
> Are there tests with the same name?
> ...

The issue is that [Uri.fsPath](https://code.visualstudio.com/api/references/vscode-api#Uri) returns a lower case drive letter instead of an upper case drive letter.

As a result, we give vitest a bad command and its internal filter will discard the test file.
Hence giving me the above error.

This PR solved the issue for me.
However, there might be a better place to put the call to sanitizeFilePath()

### DIFF
--- a/src/pure/runner.ts
+++ b/src/pure/runner.ts
@@ -89,7 +89,7 @@ export class TestRunner {
     const outputs: string[] = []
     const env = { ...process.env, ...workspaceEnv }
     let testResultFiles = [] as File[]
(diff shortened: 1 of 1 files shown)


lbm98:
Probably fixes Issue #52. Notice the line

>  DEV  v0.17.1 d:/projects/Kronos

where we have indeed have a small drive letter.

Issue #64, opened 2022-07-24
Debug specific test filter not applied on Windows


a-type:
**Describe the bug**
When trying to debug a specific test block, the debug process just runs all tests in the file on Windows with Yarn.

**To Reproduce**
Steps to reproduce the behavior on [the example project](https://github.com/zxch3n/vitest-ext-basic-example):

1. `yarn install`
2. Add a second test block
```
import { describe, it, expect} from 'vitest';

describe('add', () => {
  it ('1 + 1', () => {
    expect(1 + 1).toBe(2);
  })

  it ('2 + 2', () => {
    expect(2 + 2).toBe(4);
  })
})
```
3. Add a breakpoint on line 5, inside the "1 + 1" test
4. Open the Vitest sidebar and select "Debug test" on "2 + 2"
5. The breakpoint is hit, indicating "1 + 1" is being run

**Expected behavior**
When debugging a specific test, I expect to only run that test

**Screenshots**
![image](https://user-images.githubusercontent.com/2829772/180664850-1686efb4-a2de-4cd9-afc6-d76f065b65a0.png)
Shown above: "2 + 2" is being debugged, but the debugger has broken inside "1 + 1"

The full output for the debug test run:

```
"C:\Program Files\nodejs\node.exe" .\node_modules\vitest\vitest.mjs test D:/vitest-ext-basic-example/add.test.ts -t ""add 2 + 2"" --api 52918
 DEV  v0.12.10 d:/vitest-ext-basic-example
node_modules/vitest/dist/chunk-vite-node-externalize.464ab3dd.js:10480
      API started at http://localhost:52918
node_modules/vitest/dist/chunk-vite-node-externalize.464ab3dd.js:10480
Could not read source map for file:///d:/vitest-ext-basic-example/@vite/env: ENOENT: no such file or directory, open 'd:\vitest-ext-basic-example\@vite\env.mjs.map'
```

**Environment**

(Paste info.txt content generated by [the example project](https://github.com/zxch3n/vitest-ext-basic-example))

- OS: [e.g. macOS] Windows 11
- VSCode version: 1.69.2
- Vitest version: 0.12.10
- Vitest plugin version: v0.2.23 (prerelease)

(filled manually, the setup script does not work on Windows either)

```
Vitest settings updated
'C:\Users\a-typ\AppData\Local\Programs\Microsoft' is not recognized as an internal or external command,
operable program or batch file.
node:internal/process/promises:246
[...]

PR #69, opened 2022-08-01
fix: run partial test suite on windows (#64)

lmexo:
Removes quotes from run-configuration on windows


zxch3n:
Thanks for looking into this!

PR #78, opened 2022-08-28
feat: Changing the background color of StatusBarItem 

robertrosman:
Awesome extension @zxch3n!

Added a small tweak to change the background color of StatusBarItem to "errorBackground" when some tests are failing. An optional setting, true by default.

Of course open to feedback :)


zxch3n:
Cool! LGTM

PR #83, opened 2022-09-11
Make debug work with yarn 3 workspaces

### LINKED ISSUE #84: Debug not working with yarn 3 workspaces
**Describe the bug**
When using yarn 3 workspaces, test debugging might not work, depending on the location of the `vitest` executable.

**To Reproduce**
I opened #83 and added the `samples/multi-root-workspace-no-root` sample, and made a fix.
Without the last commit of my pr with the fix, the following error will happen when trying to debug a test:
image

**Expected behavior**
Debug should work.

adriencaccia:
Fix #84 

As explained in the comment of the second commit:

When using `yarn 3` with workspaces though, the vitest executable can be present only in the project's root `node_modules` folder.

The solution I propose is to iterate recursively on parent directories until we find the vitest executable.


seivan:
> When using `yarn 3` with workspaces though, the vitest executable can be present only in the project's root `node_modules` folder.


An alternative approach is to do what yarn itself does when it needs access to executables.
Going through pnipify is one option, another one is the same as the VSCode extension for mapping code through the archived files.

nevir:
With Yarn v4 (or v3 with pnp on), you also have a case where there is **no** `vitest` executable anywhere up the path (no node_modules folder at all)

sheremet-va:

It doesn't support pnp yet, feel free to open a separate PR to fix that.

Issue #55, opened 2022-07-14
[not support vitest@v0.18.0] test successfully, but loading infinitely.


SoonIter:
**Describe the bug**
I installed vitest V0.18.0,and it doesn`t work.

**To Reproduce**
Steps to reproduce the behavior on [the example project](https://github.com/zxch3n/vitest-ext-basic-example):
1. npx degit SoonIter/sooniter-lib-template-monorepo my-lib-name
2. run **ni**
3. go to the "test" folder
4. run this one is Ok,
<img width="749" alt="image" src="https://user-images.githubusercontent.com/79413249/178899938-4dd08faf-f74f-4b8b-a6c4-f8faeae90032.png">
but the extension works fail
<img width="939" alt="image" src="https://user-images.githubusercontent.com/79413249/178900272-129ad135-62db-4898-9add-b44142025ade.png">

**Expected behavior**
work well

**Screenshots**
<img width="757" alt="image" src="https://user-images.githubusercontent.com/79413249/178900525-92ef6558-c7d4-4ae9-860f-21ee85593f8b.png">

Maybe the problem is to run-once or run-watch.
Version 0.2.20

SoonIter:
After downgrading the vitest to @0.15.2, it works well.

beorn:
vitest 0.18.1 also doesn't work

watonyweng:
Hi @SoonIter ,Please upgrade `vitest` to the latest version.

SoonIter:
> Hi @SoonIter ,Please upgrade `vitest` to the latest version.

OK, it now works well with `vitest v0.18.1`

beorn:
The latest release, v0.2.20 (from 15 days ago), doesn't work with v0.18.1 for me.  Just get an infinite spinner.

jdk2pq:
Same issue as @beorn. I think this issue should be re-opened.

maggie9595:
> Same issue as @beorn. I think this issue should be re-opened.

same!

laozhu:
> Same issue as @beorn. I think this issue should be re-opened.

pnpm monorepo here with vitest `0.19.1`, the same problem.

geekox86:
Greetings. Same here.

I can run the tests and they pass from the terminal. However, running them from the extension gets the indicator spinning forever.

BTW, I also cloned your repo to test with your samples and the issue still exists.

macOS Big Sur 11.6.5
Version: 1.70.0-insider

"typescript": "^4.7.4",
[...]

PR #93, opened 2022-10-09
Changed the vite command parameter for the --api to --api.port instead

c00:
Related to issue https://github.com/vitest-dev/vscode/issues/55

- Changed the vite command parameter for the api to `--api.port` instead of `--api` to allow manual setting of the host. 
- Updated vitest version in the basic sample
- Added a setting to the basic sample to set the host manually
- Updated version and changelog


hiepxanh:
@zxch3n  can you take a look?

zxch3n:
Thanks!

PR #94, opened 2022-10-13
Vite 3 migration

falynx:
Various changes to migrate to Vite 3.0. Only tested on Windows 11.

- Use --api.port instead of --api.
- Bind to 127.0.0.1 instead of localhost ([breaking change](https://github.com/vitejs/vite/commit/49c0896))
- Fix parsing error by escaping special characters in test names (e.g., `"add 1 + 1"` => `"add 1 \+ 1"`)
- Stop capitalizing the first character of the test file path name. I'm not entirely sure why this was being done originally, but it was creating a mismatch with the way vitest was gathering the test file names. Vitest does a simple case-sensitive `.includes()` check, so `c:\` was not matching the `C:\` in the paths that vitest is producing. [This case-sensitive matching](https://github.com/vitest-dev/vitest/blob/9dbe4e4af7e580ebe33efe243d4137460014f3e4/packages/vitest/src/node/core.ts#L511) was breaking the "Debug Test" feature.

### DIFF
--- a/src/pure/ApiProcess.ts
+++ b/src/pure/ApiProcess.ts
@@ -94,7 +94,7 @@ export class ApiProcess {
     else {
       this.customStartProcess({
         cmd: this.vitest.cmd,
-        args: [...this.vitest.args, '--api', port.toString()],
+        args: [...this.vitest.args, '--api.port', port.toString()],
         cfg: {
           cwd,
           env: { ...process.env, ...getConfig(this.workspace).env },
@@ -135,7 +135,7 @@ export class ApiProcess {
   private _start(debouncedLog: (line: string) => void, port: number, cwd: string) {
     this.process = execWithLog(
       this.vitest.cmd,
-      [...this.vitest.args, '--api', port.toString()],
+      [...this.vitest.args, '--api.port', port.toString()],
       {
         cwd,
         env: { ...process.env, ...getConfig(this.workspace).env },
--- a/src/pure/ApiProcess.ts
+++ b/src/pure/ApiProcess.ts
@@ -94,7 +94,7 @@ export class ApiProcess {
     else {
       this.customStartProcess({
         cmd: this.vitest.cmd,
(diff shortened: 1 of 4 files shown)


hiepxanh:
that good, hope this will be merge soon

zxch3n:
Just tested it on mac. Looks great. Thanks!

hiepxanh:
@falynx  please take a look, I tested you new code but the case sensitive on window still not working, it's fail on my windows computer on version 0.2.32

image


image

```
[INFO 2:35:05 PM] 

> losa@0.0.0 test:run g:\small-cosmos\losa

> vitest run --config ./configs/vitest/vitest.config.ts "g:/small-cosmos/losa/libs/tester/src/lib/server/register.test.ts" "-t" "Register" "--api.port" "57204" "--api.host" "127.0.0.1"



[INFO 2:35:05 PM] WS Close
```

can you make a little change to update the driver to be uppercase on windows? @falynx

hiepxanh:
image
image

I get this, look like pnpm and yarn not recognize, maybe because of this.
image

@falynx  what is your environment? 
 I'm thinking of check if it's window we uppercase?

hiepxanh:
@falynx thank you, it work on the example repo, maybe it happen on my project config, I will try to figure out

hiepxanh:
I understand now, it must be happen because I'm using vitest 3x instead of 2x which have different path result which cause test fail
image

hiepxanh:
@falynx  it simply not work with vite 3, I dont know why :(

image

hiepxanh:
the basic project not work sir, I think maybe it computer or something, I try to modify it but give up, I just run vitest manually

hiepxanh:
https://drive.google.com/file/d/1CkfnobWuEH4k8qTkm5ky7HbP-9RJu_Gn/view?usp=sharing
finaly, the nodejs 18.1 is the problem

PR #111, opened 2022-12-26
fix: map test failure to correct line with latest Vitest

adrienbaron:
Happy Holidays 🎄!

Starting Vitest 0.25.x, Vitest started doing sourcemaps differently and stopped sending the parsed `stacks` as part of the api that this plugin uses to communicate with it.

As a result, test failure would be incorrectly mapped in the UI:

https://user-images.githubusercontent.com/4410247/209554471-395e0da8-fc9e-428c-90eb-8e1e71c13b34.mov

I've openned [a PR in Vitest](https://github.com/vitest-dev/vitest/pull/2563) that fixes this and sends stacks again.

However that PR in Vitest isn't enough by itself. Vitests used to include `sourcePos` (the sourcemapped position of the error) in the `ParsedStack`, but as stacks are now always source mapped, @sheremet-va pointed out this didn't make sense anymore and we removed it.

This PR catters for the case where Vitest does sends the stacks, but doesn't send `sourcePos`, which will be the case as soon as the PR on Vitest side is released.

With both the Vitest PR and this PR in, the extension is now behaving as intended and point to the line where the test failure occurs:

https://user-images.githubusercontent.com/4410247/209554709-a956e113-bd83-41fe-995e-af49eb90e070.mov

### DIFF
--- a/src/watch.ts
+++ b/src/watch.ts
@@ -279,23 +279,36 @@ export class TestWatcher extends Disposable {
   }
 }
 
+function getSourceFilepathAndLocationFromStack(stack: ParsedStack): { sourceFilepath?: string; line: number; column: number } {
+  if (stack.sourcePos) {
+    return {
+      sourceFilepath: stack.sourcePos.source?.replace(/\//g, path.sep),
+      line: stack.sourcePos.line,
+      column: stack.sourcePos.column,
+    }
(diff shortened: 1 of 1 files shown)


zxch3n:
Thanks for looking into this!

Issue #98, opened 2022-11-02
VS Code reports tests are running but nothing happens


biergit:
**Describe the bug**
VS Code tells me tests were running but then nothing

**To Reproduce**
1. Clone https://github.com/jhipster/jhipster-lite and use the devcontainer with VS Code
2. Install this plugin
3. Change setting and set `Vitest: Command Line` to `npm run vitest --`
4. Select a test and run it
5. VS Code tells me tests are running but then nothing happens. The loading gif is shown and it won't disappear.

**Environment**
mcr.microsoft.com/vscode/devcontainers/java:0-17-bullseye

**Other info**
Debugging tests fails with `No test files found, exiting with code 1`

sebasptsch:
likewise here.

The command output seems to report the tests completed successfully but the plugin doesn't resolve the result.

BePo65:
Same for me with:
Vitest version: 0.25.2
VSCode version: 1.73.1 (system setup)
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Windows_NT x64 10.0.19045

As said by others: the "output" shows the results, but the "running circles" in the test pane never stop turning.

And when I try to debug 1 test, the debugger starts, but never hits a breakpoint. Looking at the "Terminal", I find a warning that "missing dependency 'vite'". As this is not a vite project (but a pure typescript library), I installed vite as  devDependency and solved this issue; but vitest should not require an installation of vite.

PaulJNewell77:
I have the same issue with essentially the same config as above:
Version: 1.73.1 (user setup)
Commit: 6261075646f055b99068d3688932416f2346dd3b
Date: 2022-11-09T04:27:29.066Z
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Windows_NT x64 10.0.22621
Sandboxed: No

bliu13:
My issue may be tangentially related. I noticed something interesting when I looked at the output.

package.json
```
{
  "exports": "./index.js",
  "node": ">=14.16",
  "scripts": {
    "test": "vitest --watch",
    "testCoverage": "vitest run --coverage --passWithNoTests --reporter verbose"
  },
  "type": "module",
  "devDependencies": {
    "@vitest/coverage-c8": "^0.25.4",
    "c8": "^7.12.0",
    "msw": "^0.49.1",
[...]

PR #113, opened 2022-12-29
fix: tests still say Running after test run finishes

IGx89:
Was only waiting a second for vitest to start up before giving up looking for results, which is too short for slower computers. This extends it to wait up to 10 seconds, still short of the original default of 20 seconds. This likely fixes some of the issues reported in #98.

Not addressed here is the issue of no feedback if the client fails to connect, that functionality appears to be missing from vitest's client.


gregpalaci:
Still an issue

nubunto:
still an issue, still happens 2 years later, original issue closed and locked. great work!

PR #119, opened 2023-01-13
feat: make skipFiles of launch config configurable

geersch:
The launch configuration used for attaching the debugger sets `skipFiles` to `["<node_internals>/**", "**/node_modules/**"]` to make sure you are only debugging your own code. We experienced issues with the debugger being extremely slow to hit the first breakpoint. Experienced the same behaviour if we used this config with Jest. Setting it to an empty array resolved the issue for us.
Sometimes we also want to debug the `node_modules` so I made this configurable via the `debugExclude` setting, defaults to `["<node_internals>/**", "**/node_modules/**"]`.


zxch3n:
Can you use the `exclude` config instead? You can use it as the skipFiles in the debug setting

geersch:
That's meant to exclude test files, it doesn't influence `skipFiles`.
How would I be able to use this setting as `skipFiles` in de debug setting?

zxch3n:
How about renaming it to `debugExclude`

geersch:
Ok, I renamed the setting to `debugExclude`.

Issue #124, opened 2023-01-20
Vitest version not detected


asbermudez:
**Describe the bug**
After latest update (0.2.37) the extension does not allow to run tests because it does not detect the Vitest version on the workspace.
The package.json and node_modules folder are in the root of the open workspace.

**To Reproduce**
Steps to reproduce the behavior on [the example project](https://github.com/zxch3n/vitest-ext-basic-example):

1. Install latest version 0.2.37 of the extension
2. Immediately get the popup that version is under 0.12.0

**Expected behavior**
The extension should detect correctly the workspace version

**Screenshots**
Message
![image](https://user-images.githubusercontent.com/46224115/213727300-e1f2fb37-d51b-4231-936e-3f82a28a9639.png)

Version and console output
![image](https://user-images.githubusercontent.com/46224115/213727139-dd03fac6-a464-4f61-b51b-e86b9057e0bb.png)

Folder structure of workspace
![image](https://user-images.githubusercontent.com/46224115/213726928-523902a4-b678-43bb-86e7-29cfe1d86fc3.png)

**Environment**
- OS: Windows 11
- VSCode version: 1.74.2
- Vitest version: 0.27.0
- Vitest plugin version: 0.2.37

ffxsam:
Experiencing this as well.

JAKimball:
This is a duplicate of #88.

segevfiner:
I'm hitting this on macOS with pnpm

segevfiner:
It works with Vitest plugin 0.2.36

ngirardin:
@asbermudez this was probably broken by https://github.com/vitest-dev/vscode/pull/118.

Could you please try the steps at https://github.com/vitest-dev/vscode/issues/88#issuecomment-1399518132 and post your results there?

ngirardin:
You're welcome 😄

Issue #90, opened 2022-10-01
Vitest plugin executes non test files


FINDarkside:
**Describe the bug**
I cannot get the vitest plugin to work in my electron project. The vitest command works just fine, it's just this plugin that's not working.  Seems like instead of running the tests it starts to run my electron files from the dist folder even though it's ignored and doesn't even contain any tests.

**To Reproduce**
Repo to reproduce: https://github.com/FINDarkside/TLD-Save-Editor-3/tree/vitest
Just click the run tests button of the plugin

**Expected behavior**
Expect the tests to pass like they do when I run `npm run test`

**Screenshots**
Here's the logs from the failed test run:
```
[INFO 14.23.46] Tests run start
[INFO 14.23.46] [Workspace "TLD-Save-Editor-3-Vue] Run tests from workspace
[INFO 14.23.46] [Execute Vitest] C:/Users/FINDarkside/Ohjelmointi/NodeJS/TLD-Save-Editor-3-Vue/node_modules/.bin/vitest.cmd C:/Users/FINDarkside/Ohjelmointi/NodeJS/TLD-Save-Editor-3-Vue/tests/parser.spec.ts
[INFO 14.23.46] Start api process at port 63908
[INFO 14.23.46] [RUN] C:/Users/FINDarkside/Ohjelmointi/NodeJS/TLD-Save-Editor-3-Vue/node_modules/.bin/vitest.cmd C:/Users/FINDarkside/Ohjelmointi/NodeJS/TLD-Save-Editor-3-Vue/tests/parser.spec.ts
[INFO 14.23.46] [RUN.cwd] C:/Users/FINDarkside/Ohjelmointi/NodeJS/TLD-Save-Editor-3-Vue
---- Bunch of WS Close message ---
[INFO 14.23.48] WS Opened
[INFO 14.23.48] WS Opened
[INFO 14.23.48] 
 DEV  v0.23.4 C:/Users/FINDarkside/Ohjelmointi/NodeJS/TLD-Save-Editor-3-Vue
      API started at http://localhost:63908
vite v3.1.3 building for test...
watching for file changes...
build started...
[INFO 14.23.49] transforming...
✓ 1 modules transformed.
rendering chunks...
dist/electron/preload/index.js   2.14 KiB / gzip: 0.86 KiB
dist/electron/preload/index.js.map 4.01 KiB
built in 331ms.
vite v3.1.3 building for test...
watching for file changes...
build started...
transforming...
✓ 1 modules transformed.
[...]

Issue #13, opened 2022-04-06
Use closest `vitest.config.ts` from the test file for mono repo


vlechemin:
For now there isnt a way to set a configuration file for a package inside a monorepo.

https://github.com/zxch3n/vitest-explorer/issues/6#issuecomment-1090037324

Thanks!

norman-ags:
Hi! I'm also using monorepo. I have thoughts for other kind of setup (inspired by turborepo example) where:

- A monorepo that is using only vite just for test (other projects may use webpack etc)
- There will be a subpackage for only test config(vitest config)
- Other subpackages will consume the shared test config package

Not sure if this kind of setup is a good idea, here's a [sample ](https://github.com/vercel/turborepo/tree/main/examples/kitchen-sink )from turborepo, it's jest but I think it's similar.

laozhu:
The same problem here, I use turborepo for monorepo, and just some packages in it using vitest, I don't want to put vitest config to root dir.

kevinvalk:
+1

dgdolzhenko:
+1

marsidev:
Same issue here

redplant3d:
+1

luxaritas:
As a note: something like `npm exec vitest -ws -- --passWithNoTests` will work if you're only running individual test suites, but will not work for running all tests at the same time. Presumably that's because when the extension runs with a filter of files just within one particular package, `vitest` will run in each package but exit without doing anything because the filter it passes only matches files in one package. However, if that doesn't hold, as soon as the API reports results for the first package that has tests, the process is killed before vitest is run on any subsequent packages. Also worth noting that if you have many packages, running vitest against every single package when you only need to run it for one is quite slow. :)

I'd imagine the desired solution here would be something like what dbaeumer.vscode-eslint does with the `eslint.workingDirectories` setting - you can either configure (or let it determine via the presence of package.json and eslint config files) what directories are usable CWDs to launch eslint from - so if packages/mypackage is a valid working directory, when packages/mypackage/test.ts needs to be run against, packages/mypackage is picked up on as the directory that should be used.

yarinsa:
Same here turborepo & vitest, will be extremely useful, its a blocker for us from using the vscode extension

jameslounds:
+1 can't use the extension with my monorepo :/

jimmyshaka:
Add me to the camp of people who would love to use this, but can't because we use vitest in a monorepo

[...]

PR #125, opened 2023-01-22
fix: Add better version detection

ngirardin:
This is a follow-up PR for #118 which have broken some installs:
- https://github.com/vitest-dev/vscode/issues/88#issuecomment-1398074211

It seems to fix:
- https://github.com/vitest-dev/vscode/issues/124
- https://github.com/vitest-dev/vscode/issues/124
- https://github.com/vitest-dev/vscode/issues/90
- https://github.com/vitest-dev/vscode/issues/131

and probably:
- https://github.com/vitest-dev/vscode/issues/13

In some cases, vitest need to be started using `spawn(vitestCommand.cmd, [...vitestCommand.args, '-v'], envs))`, and in other by using `spawn(process.execPath, [vitestCommand.cmd, ...vitestCommand.args, '-v'], envs))`.

There are lot of OS / Node install combinations, and I don't have a clear picture of which to use and when.

The easiest fix that I could find, was to try both commands and see which one worked.

It's not the cleanest solution, but I'm open to suggestions :)

### DIFF
--- a/src/pure/utils.ts
+++ b/src/pure/utils.ts
@@ -80,31 +80,46 @@ export interface Cmd {
   args: string[]
 }
 
-export async function getVitestVersion(
-  vitestCommand?: Cmd,
-  env?: Record<string, string>,
-): Promise<string> {
-  let child
-  if (vitestCommand == null) {
-    child = spawn('npx', ['vitest', '-v'], {
-      stdio: ['ignore', 'pipe', 'pipe'],
-      env: { ...process.env, ...env },
-    })
-  }
-  else {
-    child = spawn(process.execPath, [vitestCommand.cmd, ...vitestCommand.args, '-v'], {
-      stdio: ['ignore', 'pipe', 'pipe'],
-      env: { ...process.env, ...env },
-    })
-  }
+const spawnVitestVersion = async (
+  command: string,
+  args: string[],
+  env: Record<string, string | undefined>,
+): Promise<string | undefined> => {
+  const child = spawn(command, args, {
+    stdio: ['ignore', 'pipe', 'pipe'],
+    env,
+  })
 
   // eslint-disable-next-line no-unreachable-loop
   for await (const line of chunksToLinesAsync(child.stdout)) {
     child.kill()
     return line.match(/vitest\/(\d+.\d+.\d+)/)![1]
   }
+}
+
+const getVersionOrThrow = (vitestCommand: string | undefined): string => {
+  if (vitestCommand !== undefined)
+    return vitestCommand
 
   throw new Error(`Cannot get vitest version from "${JSON.stringify(vitestCommand)}"`)
+}
+
+export async function getVitestVersion(
+  vitestCommand?: Cmd,
+  env?: Record<string, string>,
+): Promise<string> {
+  const envs = { ...process.env, ...env }
+
+  if (vitestCommand == null)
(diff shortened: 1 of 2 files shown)


ngirardin:
@zxch3n any thoughts on the fix? (I can't add you as a reviewer)

src/pure/utils.ts
```diff
@@ -80,33 +80,48 @@ export interface Cmd {
...
+  })
 
   // eslint-disable-next-line no-unreachable-loop
   for await (const line of chunksToLinesAsync(child.stdout)) {
     child.kill()
     return line.match(/vitest\/(\d+.\d+.\d+)/)![1]
   }
+}
+
+const getVersionOrThrow = (vitestCommand: string | undefined): string => {
```
zxch3n, line 100: The arg does not seem like vitestCommand

src/pure/utils.ts
```diff
@@ -80,33 +80,48 @@ export interface Cmd {
...
+  env?: Record<string, string>,
+): Promise<string> {
+  const envs = { ...process.env, ...env }
+
+  if (vitestCommand == null)
+    return getVersionOrThrow(await spawnVitestVersion('npx', ['vitest', '-v'], envs))
+
+  const version = getVersionOrThrow(await spawnVitestVersion(vitestCommand.cmd, [...vitestCommand.args, '-v'], envs))
+
+  if (version !== undefined)
```
zxch3n, line 118: It's impossible to be undefined here

zxch3n (changes requested):
This fix should use `try` and `catch` to avoid throwing in each step, right?

arimgibson:
Hey @ngirardin , let me know if you need help making those suggestions from @zxch3n ; happy to help move this forward if there's anything I can do!

ngirardin:
Thanks @zxch3n for the review.

I've made some changes to the code:
- log the commands that are spawned
- added a message to invite users to open an issue and provide their logs if no version is detected
- trying both of the spawn methods for starting `npx` (when `command == null`)
- added some tests (WIP needs to mock spawn to test all the cases)
- added a check when doing the `line.match` to avoid crashing if the version is not found in the output

@arimgibson, sorry for the delay, please let me know if you see something to change on this code :)

src/pure/utils.ts
```diff
@@ -80,31 +81,73 @@ export interface Cmd {
...
+    stdio: ['ignore', 'pipe', 'pipe'],
+    env,
+  })
+
+  child.on('error', () => {
+    log.info('Command not found')
+  })
 
   // eslint-disable-next-line no-unreachable-loop
   for await (const line of chunksToLinesAsync(child.stdout)) {
```
zxch3n, line 106: Maybe you should use try catch to wrap 106~114 here? Otherwise, it may just throw on 127 or 133 line

ngirardin: Is `chunksToLinesAsync` throwing if the `spawn` fails?

When testing locally, if the spawn fails, we're not entering in the `for` loop and `detectVitestVersion` run the second command.

zxch3n: I was afraid of the for await might throw. If you've tested it, that's fine. Thanks!

ngirardin: Thank you for the review and the merge.

I'll add some tests, so we'll be able to test the edge cases if some new bugs are open, I found a lib to mock `spawn`

Issue #152, opened 2023-04-27
Bug "NaN%, 2 skipped"


JoaoVictor6:
**Describe the bug**
When not exists test, the status on bottom bar print a NaN%
![image](https://user-images.githubusercontent.com/68869379/235004515-7cdd37de-ee96-425c-8fa7-92d062114b2d.png)

**To Reproduce**
Steps to reproduce the behavior on [the example project](https://github.com/zxch3n/vitest-ext-basic-example):

1. Create a basic suite test.
```ts
describe('test', () => {
  it('should be pass', () => {
    expect(true).toBeTruthy()
  })
})
```
2. Run only this suite
```ts
describe.only('test', () => {
  it('should be pass', () => {
    expect(true).toBeTruthy()
  })
})
```
3. Now, skip the test
```ts
describe.only('test', () => {
  it.skip('should be pass', () => {
    expect(true).toBeTruthy()
  })
})
```
![image](https://user-images.githubusercontent.com/68869379/235005194-d2dd0a4b-2830-45a8-9eb6-d7777b5198b3.png)

**Expected behavior**
I believe that in this case, it would be interesting to show something as: "0%, 2 skipped"

- OS: Manjaro Linux
- VSCode version: 1.76.2
- Vitest version: ^0.30.1
- Vitest plugin version: v0.2.39

JoaoVictor6:
I fix this issues :)
![image](https://user-images.githubusercontent.com/68869379/235011837-77bd2653-2620-4d6c-8cd0-05ca2bdf9147.png)

PR #153, opened 2023-04-27
fix "NaN%, 2 skipped" bug

JoaoVictor6:
This PR resolves the #152 issue

### DIFF
--- a/src/StatusBarItem.ts
+++ b/src/StatusBarItem.ts
@@ -37,9 +37,14 @@ export class StatusBarItem extends vscode.Disposable {
   ) {
     this.item.command = Command.StopWatching
     const total = passed + failed
+    const percentOfExecutedTests = Number((passed / total * 100).toFixed(0))
+    const percentIsValid = Boolean(percentOfExecutedTests)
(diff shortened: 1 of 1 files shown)


src/StatusBarItem.ts
```diff
@@ -37,9 +37,14 @@ export class StatusBarItem extends vscode.Disposable {
   ) {
     this.item.command = Command.StopWatching
     const total = passed + failed
+    const percentOfExecutedTests = Number((passed / total * 100).toFixed(0))
+    const percentIsValid = Boolean(percentOfExecutedTests)
```
zxch3n, line 41: percentOfExecutedTests might be 0, and this will be false

JoaoVictor6: I used a boolean because if `percentOfExecutedTests` is 0(falsy) the percent number receives a 0 of the ternary:
`const percentNumber = percentIsValid ? percentOfExecutedTests : 0﻿﻿`

However, make more sense to use an `isNaN` instead `Boolean` :) Thanks for the review!

PR #174, opened 2023-07-18
feat(ci): add ci and release workflows

watonyweng:


ffMathy:
This was implemented over in #239 instead.

Issue #148, opened 2023-04-02
Running vitest from status bar changes workspace setting `"testing.automaticallyOpenPeekView": "never"`


sagarpanchal:
**Describe the bug**
it should not modify workspace settings

**To Reproduce**
click on vitest on statusbar

1. check workspace `settings.json`

**Expected behavior**
it should not modify workspace settings

langaads:
Same behavior for me, it's annoying, i keep seeing people commiting this change by accident in our workspace settings.

huwshimi:
I've created [a PR](https://github.com/vitest-dev/vscode/pull/188) so that settings.json is no longer modified, and users who wish to have the peek view in watch mode can set `"testing.automaticallyOpenPeekViewDuringAutoRun": true`.

PR #188, opened 2023-10-19
Run tests in continuous mode

huwshimi:
## Done

Run tests in continuous mode so the peek view can be controlled by the settings.

Fixes: #148.

## QA

- Run this extension.
- Open a test file and change a test to fail.
- Run the failing test and the peek view should appear.
- Close the peek view and then click the Vitest item in the status bar to run the tests in watch mode.
- The tests should run but the peek view should not appear.
- Open settings.json and add `"testing.automaticallyOpenPeekViewDuringAutoRun": true`.
- Run this extension again.
- Open the file with the failing test.
- Click the Vitest item in the status bar to run the tests in watch mode.
- The peek view should appear when the test fails.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "url": "https://github.com/vitest-dev/vscode/issues"
   },
   "engines": {
-    "vscode": "^1.63.0"
+    "vscode": "^1.77.0"
   },
   "categories": [
     "Testing"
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "url": "https://github.com/vitest-dev/vscode/issues"
   },
   "engines": {
-    "vscode": "^1.63.0"
+    "vscode": "^1.77.0"
```
huwshimi, line 25: The continuous mode setting was introduced in 1.77.0

src/discover.ts
```diff
@@ -112,6 +112,7 @@ export class TestFileDiscoverer extends vscode.Disposable {
             workspaceFolder.uri,
             include,
           )
+          const workspacePath = workspaceFolder.uri.fsPath
```
huwshimi, line 115: This is a drive-by fix as this was showing an error due to the missing variable when the refresh button was clicked.

Issue #151, opened 2023-04-26
Support `vitest.rootPath` option


mizdra:
**Is your feature request related to a problem? Please describe.**
I want to use this extension even in cases where `vitest.config.ts` is not directly under the workspace root directory.

For example, imagine a workspace with the following directory structure.

```
workspace-root/
├ .github/
├ backend/
├ ui/
│ ├ node_modules/
│ ├ src/
│ │ ├ index.ts
│ │ └ index.test.ts
│ ├ package.json
│ └ vitest.config.ts
└ terraform/
```

In this workspace, backend, frontend, and IaC code are clearly separated by directory. vitest is used only for frontend code, so `vitest.config.ts` is under `ui/` directory. Therefore, the extension must run vitest with `ui/` directory as the current directory, not the workspace root directory.

However, this extension does not have an option to change the current directory where vitest runs.

**Describe the solution you'd like**
Add the `vitest.rootPath` option; vscode-jest has a similar option: https://github.com/jest-community/vscode-jest#rootpath

**Describe alternatives you've considered**
Nothing.

**Additional context**
Nothing.

eomeara:
I have same problem and same desired fix.

sheremet-va:
This should be fixed in pre-release 0.5.0 and higher. Note that the extension now requires Vitest 1.4.0 or higher.

The extension now automatically finds the config and uses it as the working directory.

Issue #49, opened 2022-06-21
Nested package structure


timges:
**Is your feature request related to a problem? Please describe.**

When you have a nested pkg structure like root/frontend/... the extension still tries to find your package.json at root/package.json instead of root/frontend/package.json

**Describe the solution you'd like**
Make the root path customizable

Inviz:
Having the same issue. Couldn't solve it. Some `cwd` option would really help.

ngirardin:
Hi guys, that should be solved with the latest version (v0.2.39).

Could you give it a try?

sheremet-va:
This should be fixed in pre-release 0.5.0 and higher. Note that the extension now requires Vitest 1.4.0 or higher.

PR #190, opened 2023-10-20
Support root config

huwshimi:
## Done

- Add `rootPath` to support projects that are not at the root of the workspace (fixes: #151, fixes: #49).
- Support the Vitest `root` config for tests that are not at the root of a project.

## QA

- Run this extension with the Nested Root Sample project.
- It should find and run the `add.test.ts` tests.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,6 +16,16 @@
       "outFiles": ["${workspaceFolder}/out/**/*.js"]
     },
     {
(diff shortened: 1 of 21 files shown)


ffMathy:
Closing in favor of using `vitest.workspace.ts` at the root-level. 

See more here: https://vitest.dev/guide/workspace

Let me know if you disagree with this decision.

PR #192, opened 2023-10-28
fix: exclude/include config handling and missing workspacePath

mash-graz:
This PR fixes two issues related to test file inclusion/exclusion:

1. in `getCombinedConfig()` the deprecated vscode settings or better its default values, when  not found, always overruled the settings in the vite and vitest configs. By this change the entries in both configuration variants get merged.
 
Fixes Bugs in:  #187 

2. The fix in `discover.ts` just adds an obviously missing variable, which caused an error
 
Pasted image


mash-graz:
> What is required to get this merged?

I'm also somehow surprised, that it seems to be completely ignored by the maintainers (@zxch3n) and developers of the patched code (@huwshimi).

PR #199, opened 2023-12-08
chore(settings): add "e2e" to default exclude list

DrJume:
`e2e` is a general folder name used for End-To-End tests, like `cypress`.


ennioVisco:
This makes a lot of sense!

ffMathy:
This is no longer needed, as exclude and include have been fully removed. The vitest config is now being used instead.

PR #213, opened 2024-01-17
dev: adapt to vitest@1.2

keroxp:
This PR includes an adaption to the latest vitest distribution v1.2.2 (2024-02-16)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,11 +2,21 @@ import * as vscode from 'vscode'
 import semver from 'semver'
 import type { WorkspaceConfiguration, WorkspaceFolder } from 'vscode'
 import type { ResolvedConfig } from 'vitest'
-import { configDefaults } from 'vitest/config'
 import { isDefinitelyVitestEnv, mayBeVitestEnv } from './pure/isVitestEnv'
 import { getVitestCommand, getVitestVersion, isNodeAvailable } from './pure/utils'
 import { log } from './log'
 export const extensionId = 'zxch3n.vitest-explorer'
+
+// Copied from https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/defaults.ts
+// "import { configDefaults } from 'vitest'" throws unexpected URL error
+const defaultInclude = ['**/*.{test,spec}.?(c|m)[jt]s?(x)']
+const defaultExclude = [
+  '**/node_modules/**',
+  '**/dist/**',
+  '**/cypress/**',
+  '**/.{idea,git,cache,output,temp}/**',
+  '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build,eslint,prettier}.config.*',
+]
 
 export function getConfigValue<T>(
   rootConfig: WorkspaceConfiguration,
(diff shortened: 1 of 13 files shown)


src/config.ts
```diff
@@ -2,12 +2,22 @@ import * as vscode from 'vscode'
...
 import type { WorkspaceConfiguration, WorkspaceFolder } from 'vscode'
 import type { ResolvedConfig } from 'vitest'
-import { configDefaults } from 'vitest/config'
 import { isDefinitelyVitestEnv, mayBeVitestEnv } from './pure/isVitestEnv'
 import { getVitestCommand, getVitestVersion, isNodeAvailable } from './pure/utils'
 import { log } from './log'
 export const extensionId = 'zxch3n.vitest-explorer'
 
+// Copied from https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/defaults.ts
+// "import { configDefaults } from 'vitest'" throws unexpected URL error
```
keroxp, line 11: This statement throws "Invalid URL" error during activation (from `node_modules/vite/dist/node-cjs/publicUtils.cjs`.  That file includes codes like below:

```js
const { version: version$2 } = JSON.parse(fs$1.readFileSync(new URL('../../package.json', (typeof document === 'undefined' ? require('u' + 'rl').pathToFileURL(__filename).href : (_documentCurrentScript && _documentCurrentScript.src || new URL('node-cjs/publicUtils.cjs', document.baseURI).href)))).toString());
```
It won't work in the bundled `dist/extension.js`. I couldn't find any solution and gave up. Are anyone knows something?

keroxp:
@zxch3n Hi, would you have time to see this PR?

MilanKovacic:
I am also interested to hear if you are still interested in maintaining the project. I did some updates in https://github.com/MilanKovacic/vitest-visual-studio-code-extension

src/config.ts
sheremet-va: I would expect Vite to not be bundled with this extension, maybe it can be externalized/tree-shaken?

src/watch.ts
```diff
@@ -414,14 +406,13 @@ export function syncTestStatusToVsCode(
...
-    type: 'suite' | 'test',
   ): TestDescribe | TestCase {
     let ans: (TestDescribe | TestCase) | undefined
     for (const candidate of candidates) {
-      if (type === 'suite' && !(candidate instanceof TestDescribe))
+      if (task.type === 'suite' && !(candidate instanceof TestDescribe))
         continue
 
-      if (type === 'test' && !(candidate instanceof TestCase))
+      if (task.type === 'test' && !(candidate instanceof TestCase))
```
sheremet-va, line 415: What about type `custom`?

sheremet-va (changes requested):
Looks good! Just a few questions

src/config.ts
sheremet-va: `vitest/config` doesn't actually import anything from Vite, so I'm not sure why it even gets bundled here 🤔

hi-ogawa: It looks like tsup/esbuild's default tree-shaking is not doing much and keeps unused `require("vite")`. Actually tsup supports rollup-based tree-shaking https://tsup.egoist.dev/#tree-shaking, so this config seems to work but rebuild slows down a bit.

```ts
// tsup.config.ts
import { defineConfig } from "tsup"

export default defineConfig({
  entry: ["./src/extension.ts"],
  external: ["vscode", "vite"],
  treeshake: true,
})
```

Personally, just inlining `defaultInclude` and `defaultExclude` is equally fine as well.

sheremet-va: tsup also supports plugins where we can manually remove vite import

hi-ogawa: Do you mean esbuild plugins?

sheremet-va: Yes

src/watch.ts
keroxp: I don't understand what's type='custom'. It should be treated as which one? It seems to be newly introduced in v1.

src/config.ts
keroxp: It is ideal to import the original config from vitest. I think it's problem of vitest bundle. `new URL` with relative path resolution won't work imported bundle. Anyway, can it be acceptable?

sheremet-va: Ok, let's keep hardcoded config for now

Issue #7, opened 2022-03-30
describe.each and test.each are shown as 'test result not found'


lampewebdev:
Hello,

when usuing `.each` the extension can not find the tests.

<img width="2136" alt="image" src="https://user-images.githubusercontent.com/44268468/160782371-699e7f05-a353-42fc-b146-d5e147e97805.png">
<img width="1477" alt="image" src="https://user-images.githubusercontent.com/44268468/160782549-ed7b3590-d25d-4d63-aef7-742f32e465d9.png">

zxch3n:
`.each` is not supported yet. I'll find a way to make it work

lampewebdev:
@zxch3n thanks for the info.

Are you accepting PR's?

If you could just point me to the right file, I could take a look at it :)

zxch3n:
PRs are welcome! But I'm afraid this might not be an easy fix. Because we cannot infer the dynamic test name with 100% correctness from the static text content (which is how this plugin works right now). 

There are two options left

- Only register the dynamic tests' ancestors, and aggregate the test results to them
- Start a Vitest server and communicate with it

I'll work on the latter in a week or two, which can also provide watch mode #4 

If you are interested you can take a look at `discover.ts` and the VSCode testing API document.

sheremet-va:
> * Start a Vitest server and communicate with it

Yeah, I think Vitest has an api to communicate with `--ui`? Can you use that?

Nisgrak:
Maybe this is working now? In my repo the test shows the actual results

watonyweng:
Hi @lampewebdev, the latest version of `vitest` already supports `describe.each` and `test.each`, does the test command still prompt `test result not found`?

unikitty37:
Using the example from the [describe.each documentation](https://vitest.dev/api/#describe-each), I'm still only seeing placeholders rather than the expanded tests.

![image](https://user-images.githubusercontent.com/139183/225040561-a04d577b-e50c-4f4b-8ab1-0dad356a8be1.png)

Version: 1.76.1
Commit: 5e805b79fcb6ba4c2d23712967df89a089da575b
Date: 2023-03-08T16:48:08.231Z
Electron: 19.1.11
Chromium: 102.0.5005.196
Node.js: 16.14.2
V8: 10.2.154.26-electron.0
OS: Darwin arm64 21.6.0
Sandboxed: No

Vitest extension v0.2.39

wadefletch:
+1

[...]

PR #214, opened 2024-01-19
feat: support .each suite execution by regex

keroxp:
Support an execution of `.each` describe/test suite that includes `%` placeholders. vitest `--testNamePattern` opts doesn't recognize pattern with`%` placeholders. So we need to transform the given suite name by replacing `%` to `.+` regexp.

The base branch of this PR is #213.


MilanKovacic:
Fixes https://github.com/vitest-dev/vscode/issues/7.

keroxp:
Additionally, Incomplete edge cases still remain. Similar describe/test cases will be matched by the same transformed regular expression. Like:

```
(1) test.each([...])("my test %s)(() => {})
(2) test.each([...])("my test will be %s")(()=>{})
```

(2) is also executed by executing (1) and got no result, showing error, because (1) goes to `my test .+?`.

sheremet-va:
Thank you for your contribution! Can you resolve conflicts, when you have the time, please? 🙏🏻

keroxp:
Backed to WIP, still refactoring.

ffMathy (approved):
LGTM! I love this. Initially I was skeptical of the use of regex, but after seeing the implementation, it should be stable.

ffMathy:
@keroxp, @sheremet-va wants to know if we can merge this?

sheremet-va (approved):

ffMathy:
This introduces a bug. When there are more describes with the same name, you can't run the tests in the last describe.

Example repro:

```typescript
import { describe, expect, it } from 'vitest'
import { add } from '../src/add'

describe('duplicate describe', () => {
  it('test 1', () => {
    expect(add(1, 1)).toBe(2)
  })
})

describe('duplicate describe', () => {
  it('test 2', () => {
    expect(add(1, 1)).toBe(2)
  })
})
```

With the above, "test 2" can't be run and will give an error. This worked before.

ffMathy:
cc @keroxp. Got an idea for a fix? I tried, but can't understand the code.

Issue #227, opened 2024-02-15
[DevEx] Debugging of extension does not work due to missing outFiles configuration


MilanKovacic:

ffMathy:
This has been fixed by #230.

PR #230, opened 2024-02-17
fix: debugging not hitting breakpoints

ffMathy:
Breakpoints are not hit because VS Code is looking at the wrong directory for output files. In addition, there was an issue with the `--api.host` being set twice for the basic configuration.


MilanKovacic:
Fixes #227.

ffMathy:
An alternative fix could be to modify the `tsup` CLI call to be explicit about the output directory, thus respecting the `out` directory still. Up to you.

MilanKovacic (commented):
LGTM

PR #231, opened 2024-02-17
feat: nx support

ffMathy:
Builds on top of #230, so that should probably be merged first.

I had some issues getting NX working, so this PR adds a new NX sample.

The issue I was having turned out to be that when Vite runs within the sample directory, it tries to traverse upwards in directories and then would catch the root repo's `vite.config.ts` file, which would have an `exclude` section that would exclude `**/src/**` and others.

Now the global vite config has been renamed to avoid it accidentally being picked up in the samples.


ffMathy:
This is also a big source of bad reviews: https://marketplace.visualstudio.com/items?itemName=ZixuanChen.vitest-explorer&ssr=false#review-details

sheremet-va (approved):

chriswheeldon-peakon:
@ffMathy is it intended that with these changes the working directory is the containing package rather than the monorepo root?

ffMathy:
You can use a vitest.workspace.ts file and it should work fine from the NX monorepo root.

But there are some known bugs right now that are being fixed in the next release in the coming days.

Also note that the current extension has been deprecated and there is a new link to the new extension.

chriswheeldon-peakon:
> You can use a vitest.workspace.ts file and it should work fine from the NX monorepo root.
> 
> But there are some known bugs right now that are being fixed in the next release in the coming days.
> 
> Also note that the current extension has been deprecated and there is a new link to the new extension.

Thanks @ffMathy. I'm using v0.3.0 of https://marketplace.visualstudio.com/items?itemName=vitest.explorer which I think has this change in it. Using the new sample, `monorepo-nx`, I find that the working directory is the repository root during test execution, whereas running the test task with nx will change the working directory to each package's root. Is that inconsistency expected, and if so is there any plan to address it or should we look for workarounds?

ffMathy:
0.3.0 doesn't have the bugfixes that are incoming.

Hmmm I didn't know about the working directory thing. Perhaps I should make a fix for that and create another test for that scenario.

Before my change, NX tests wouldn't run at all. So at least there's progress.

Can you file an issue with the working directory problem?

chriswheeldon-peakon:
> 0.3.0 doesn't have the bugfixes that are incoming.
> 
> Hmmm I didn't know about the working directory thing. Perhaps I should make a fix for that and create another test for that scenario.
> 
> Before my change, NX tests wouldn't run at all. So at least there's progress.
> 
> Can you file an issue with the working directory problem?

Will do! Thank you!

PR #232, opened 2024-02-17
feat: support using keyword

ffMathy:
Today, if a test has the `using` keyword in it (a new TypeScript feature), Vitest won't run or discover that particular test. This fixes that problem.

### DIFF
--- a/src/pure/parsers/helper.ts
+++ b/src/pure/parsers/helper.ts
@@ -29,6 +29,7 @@ const commonPlugins: ParserPlugin[] = [
   'optionalChaining',
   'partialApplication',
   'throwExpressions',
+  'explicitResourceManagement',
   'topLevelAwait',
   ['pipelineOperator', { proposal: 'smart' }],
 ]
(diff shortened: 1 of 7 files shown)


MilanKovacic:
Can we add the "using" keyword to one of the "samples" test cases?

src/pure/parsers/helper.ts
```diff
@@ -29,6 +29,7 @@ const commonPlugins: ParserPlugin[] = [
   'optionalChaining',
   'partialApplication',
   'throwExpressions',
+  'explicitResourceManagement',
```
MilanKovacic, line 32: Can we add a comment specifying that this relates to "using" keyword?

ffMathy: Sure! Done! 👍

ffMathy:
> Can we add the "using" keyword to one of the "samples" test cases?

Done! I also added an actual unit test to the parser which verifies that it is working.

sheremet-va (approved):

PR #234, opened 2024-02-17
feat: discover on startup

ffMathy:
This PR makes it so that the Vitest extension automatically discovers tests on startup. This an explicit action today.


MilanKovacic (commented):
LGTM

PR #238, opened 2024-02-18
feat: auto watch on startup setting

ffMathy:
This PR adds a new setting that allows for Vite to start watching on startup.


ffMathy:
cc @sheremet-va

PR #246, opened 2024-02-18
ci: vsix publish

### LINKED ISSUE #141: Publish updates to Open VSX Registry
**Is your feature request related to a problem? Please describe.**
I cannot automatically update the Vitest extension from [VSCodium](https://vscodium.com/).

**Describe the solution you'd like**
Publish each update to the Vitest extension to the [Open VSX Registry](https://open-vsx.org/), so VSCodium users can access it.

**Describe alternatives you've considered**
Manually installing the Vitest extension from the [Visual Studio Marketplace](https://marketplace.visualstudio.com/items?itemName=ZixuanChen.vitest-explorer) every time it is updated.

**Additional context**
See #43, which was closed without providing a structural solution to the problem.

ffMathy:
Fixes #141


.github/workflows/build.yml
```diff
@@ -28,3 +28,32 @@ jobs:
...
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: pnpm
+
+      - run: pnpm install --frozen-lockfile
+      
+      - name: Publish to Open VSX Registry
+        uses: HaaLeo/publish-vscode-extension@v1
+        with:
+          pat: ${{ secrets.OPEN_VSX_TOKEN }}
```
hi-ogawa, line 53: Probably this first one needs ``dependencies: false` to avoid vsce/pnpm issue when building.
https://github.com/HaaLeo/publish-vscode-extension/tree/master?tab=readme-ov-file#input-parameters

Then for 2nd one, `.vsix` is already built during the 1st one, so we can use `extensionFile: ${{ steps.publishToOpenVSX.outputs.vsixPath }}` to reuse it.
https://github.com/HaaLeo/publish-vscode-extension?tab=readme-ov-file#example

ffMathy: Fixed! ❤️

hi-ogawa (commented):
Thanks!

We cannot actually test the whole thing, but I guess we could still test out packaging part without publishing by something like this:

```yml
# remove "tags" check temporary to trigger CI on this PR itself
on:
  push:
    # tags:
    #   - v*

      # ...
      # use `dryRun` and dummy `pat` to test just packaging part without publish
      - name: Publish to Open VSX Registry
        id: publishToOpenVSX
        uses: HaaLeo/publish-vscode-extension@v1
        with:
          # pat: ${{ secrets.OPEN_VSX_TOKEN }}
          pat: dummy
          dependencies: false
          dryRun: true

      - name: Publish to Visual Studio Marketplace
        uses: HaaLeo/publish-vscode-extension@v1
        with:
          # pat: ${{ secrets.VS_MARKETPLACE_TOKEN }}
          pat: dummy
          registryUrl: https://marketplace.visualstudio.com
          extensionFile: ${{ steps.publishToOpenVSX.outputs.vsixPath }}
          dryRun: true
```

Maybe worth a try?

ffMathy:
> Thanks!
> 
> We cannot actually test the whole thing, but I guess we could still test out packaging part without publishing by something like this:
> 
> ```yaml
> # remove "tags" check temporary to trigger CI on this PR itself
> on:
>   push:
>     # tags:
>     #   - v*
> 
>       # ...
>       # use `dryRun` and dummy `pat` to test just packaging part without publish
>       - name: Publish to Open VSX Registry
>         id: publishToOpenVSX
>         uses: HaaLeo/publish-vscode-extension@v1
>         with:
>           # pat: ${{ secrets.OPEN_VSX_TOKEN }}
>           pat: dummy
>           dependencies: false
>           dryRun: true
> 
>       - name: Publish to Visual Studio Marketplace
>         uses: HaaLeo/publish-vscode-extension@v1
>         with:
>           # pat: ${{ secrets.VS_MARKETPLACE_TOKEN }}
>           pat: dummy
>           registryUrl: https://marketplace.visualstudio.com
>           extensionFile: ${{ steps.publishToOpenVSX.outputs.vsixPath }}
>           dryRun: true
> ```
> 
> Maybe worth a try?

Great idea. Will try it out.

ffMathy:
Hmmm doesn't look like GitHub wants to run the new build script from a PR. Makes sense from a security perspective.

I am going to revert that commit, and then I think we just have to test on main.

hi-ogawa (approved):
Ah, maybe it's possible that CI added by non-admin doesn't run unless it's committed by admin to protect secret leak or something.

I guess it should be alright and @sheremet-va can tweak it later even if it fails.

ffMathy:
> Ah, maybe it's possible that CI added by non-admin doesn't run unless it's committed by admin to protect secret leak or something.
> 
> I guess it should be alright and @sheremet-va can tweak it later even if it fails.

Yeah indeed.

sheremet-va:
> Ah, maybe it's possible that CI added by non-admin doesn't run unless it's committed by admin to protect secret leak or something.

Yes, you need special permission to use the "Release" environment.

ffMathy:
@sheremet-va should we then merge this, and then you folks can test directly on main until it works?

sheremet-va:
> @sheremet-va should we then merge this, and then you folks can test directly on main until it works?

We can't merge this because there is a yarnlock file, and we use pnpm now.

Issue #17, opened 2022-04-14
Any chance of support for in-source testing


Aaron-Pool:
I noticed the extension doesn't pick up on [in-source tests](https://vitest.dev/guide/features.html#in-source-testing). Any chance this feature will be supported in the future. I might be willing to contribute to a PR if I new it wouldn't sit dead in the PR board indefinitely.

watonyweng:
Hi @zxch3n, if we want to support  [In-source Testing](https://vitest.dev/guide/in-source.html) in VSCode, we should add options to configure `includeSource` and dedect some files when there is a `if (import.meta.vitest){ } ` block in source code.

szulcus:
What should I do to make the [plugin ](https://marketplace.visualstudio.com/items?itemName=ZixuanChen.vitest-explorer) for VS Code work?
![image](https://user-images.githubusercontent.com/48328739/214594324-1707cefb-ce23-45aa-aedc-da7d81aa7ee4.png)

nicolasreszka:
Hello, did you find a solution ?

PR #250, opened 2024-02-19
fix: include pattern not supported

ffMathy:


package.json
```diff
@@ -153,13 +153,19 @@
     "lint": "eslint --cache .",
     "lint:fix": "eslint --cache --fix ."
   },
+  "dependencies": {
```
sheremet-va, line 156: Please, don't add dependencies to devDependencies, they won't be bundled otherwise

src/discover.ts
```diff
@@ -107,35 +98,24 @@ export class TestFileDiscoverer extends vscode.Disposable {
 
     await Promise.all(
       vscode.workspace.workspaceFolders.map(async (workspaceFolder) => {
-        const workspacePath = workspaceFolder.uri.fsPath
-        const exclude = this.config.exclude
-        for (const include of this.config.include) {
-          const pattern = new vscode.RelativePattern(
-            workspaceFolder.uri,
-            include,
+        const files = await globFiles(this.config.include, this.config.exclude, workspaceFolder.uri.fsPath)
```
sheremet-va, line 101: Should use use `this.config.dir || workspaceFolder.uri.fsPath` for better performance?

src/vscodeUtils.ts
```diff
@@ -17,10 +20,36 @@ export async function getContentFromFilesystem(uri: Uri) {
...
+ */
+export async function shouldIncludeFile(path: string, config: ResolvedConfig): Promise<boolean> {
+  const relativeId = relative(config.dir || config.root, path)
+  if (micromatch.isMatch(relativeId, config.exclude))
+    return false
+
+  if (micromatch.isMatch(relativeId, config.include))
+    return true
+
+  if (config.includeSource?.length && micromatch.isMatch(relativeId, config.includeSource)) {
```
sheremet-va, line 34: I don't think source tests are supported yet (#17), let's remove this check for now

sheremet-va (changes requested):

src/vscodeUtils.ts
ffMathy: But actually, having this check here, it would be solved. So why not? The test would display in the test explorer, which would lead to it being runnable. If you want, I can check if it is supported?

src/discover.ts
ffMathy: Great idea.

src/vscodeUtils.ts
ffMathy: Ah okay. It seemed to work when opening a new insource test file, but I think that is just on the tab open event. I'll definitely remove it then.

package.json
ffMathy: Wait, I never added them to dev-dependencies. They are added as real dependencies.

sheremet-va: I wanted to write in the first comment "don't add to "dependencies", use "devDependencies""

ffMathy: Got it. I applied it now.

src/discover.ts
sheremet-va: You marked this as resolved, but it's not

```suggestion
        const files = await globFiles(this.config.include, this.config.exclude, this.config.dir || workspaceFolder.uri.fsPath)
```

src/discover.ts
```diff
@@ -53,49 +52,41 @@ export class TestFileDiscoverer extends vscode.Disposable {
...
-            include,
-          )
-          const workspacePath = workspaceFolder.uri.fsPath
-          const watcher = vscode.workspace.createFileSystemWatcher(pattern)
-          const filter = (v: vscode.Uri) =>
-            exclude.every(x => !minimatch(path.relative(workspacePath, v.fsPath), x, { dot: true }))
-          watcher.onDidCreate(
-            uri => filter(uri) && this.getOrCreateFile(controller, uri),
+        const workspacePath = workspaceFolder.uri.fsPath
+        const files = await globFiles(this.config.include, this.config.exclude, workspaceFolder.uri.fsPath)
```
sheremet-va, line 56:
```suggestion
        const files = await globFiles(this.config.include, this.config.exclude, this.config.dir || workspacePath)
```

sheremet-va (changes requested):

src/discover.ts
ffMathy: Sorry - now I got it! :)

Issue #196, opened 2023-11-08
Vitest extension completely non functional on mac


benwainwrightawaze:
**Describe the bug**
Vitest completely fails to work on any of my vitest repos

**To Reproduce**
Steps to reproduce the behavior on [the example project](https://github.com/benwainwright/vitest-vscode-issue-repro):

1. Open Vscode
2. Install Vitest extension
3. Go to 'testing' sidebar

**Expected behavior**
I see tests that I can execute

**Actual behavior**
I see 'install additional test extensions', and no tests.

Note, the vitest output window appears to be showing websocket connections dropping:

```
[INFO 17:42:27] Trying to get vitest version from /Users/ben.wainwright/personal-repos/vitest-basic/node_modules/.bin/vitest -v...
[INFO 17:42:28] vitest/0.34.6 darwin-arm64 node-v20.6.0

[INFO 17:42:28] Vitest Workspace [vitest-basic]: Vitest version = 0.34.6
[INFO 17:42:28] WS Close
[INFO 17:42:28] RECONNECT
[INFO 17:42:28] WS Close
[INFO 17:42:29] RECONNECT
[INFO 17:42:29] WS Close
[INFO 17:42:29] RECONNECT
[INFO 17:42:29] WS Close
[INFO 17:42:30] RECONNECT
[INFO 17:42:30] WS Close
[INFO 17:42:30] RECONNECT
[INFO 17:42:30] WS Close
[INFO 17:42:31] RECONNECT
[INFO 17:42:31] WS Close
[INFO 17:42:31] RECONNECT
[INFO 17:42:31] WS Close
[INFO 17:42:32] RECONNECT
[INFO 17:42:32] WS Close
[INFO 17:42:32] RECONNECT
[INFO 17:42:32] WS Close
[INFO 17:42:33] RECONNECT
[INFO 17:42:33] WS Close
[INFO 17:42:33] RECONNECT
[INFO 17:42:33] WS Close
[INFO 17:42:34] RECONNECT
[INFO 17:42:34] WS Close
[INFO 17:42:34] RECONNECT
[INFO 17:42:34] WS Close
[INFO 17:42:35] RECONNECT
[INFO 17:42:35] WS Close
[INFO 17:42:35] RECONNECT
[INFO 17:42:35] WS Close
[INFO 17:42:36] RECONNECT
[INFO 17:42:36] WS Close
[INFO 17:42:36] RECONNECT
[INFO 17:42:36] WS Close
[...]

PR #255, opened 2024-02-20
Add support for Node.js modules in .mjs (.mts) format

egor-xyz:
This pull request introduces support for .mjs and .mts file formats for configuration files within the Vitest VSCode extension. Previously, the extension lacked the capability to recognize or utilize configuration files written in module formats. This limitation posed a significant inconvenience for developers working in codebases that prefer these formats for their ES module compatibility and other benefits.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,9 +34,13 @@
   },
   "activationEvents": [
     "workspaceContains:**/vitest.config.js",
+    "workspaceContains:**/vitest.config.mjs",
     "workspaceContains:**/vite.config.js",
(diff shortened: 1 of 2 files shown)


egor-xyz:
I'm closing PR. Looks like the extension is broken on macOS. Revert to version 0.2.42 helped. Issue: #196

egor-xyz:
I'm closing PR. Looks like the extension is broken on macOS. Revert to version 0.2.42 helped. Issue: #196

Issue #249, opened 2024-02-19
Add tests


sheremet-va:
Currently, we only have unit tests for some of the functionality, but it would be nice to also have integration tests using Electron API: https://code.visualstudio.com/api/working-with-extensions/testing-extension

MilanKovacic:
I can take this one.

ffMathy:
Sorry, didn't see this issue. I already have a PR that works for this. See the referenced PR.

MilanKovacic:
No problem, I will add a smoke end-to-end test instead.

sheremet-va:
I added a "needs test" label. Anyone who wants to help, can look through all issues with that label and write a test 😄

PR #262, opened 2024-02-23
feat: switch to vs code tests

ffMathy:
This switches to VS Code tests. Unfortunately this has to ditch Vitest in favor of Mocha, which is a requirement for this to work.

But the good thing is that this allows for real integration tests of the extension! They also run super fast, and it will make stabilizing the extension a lot easier.


package.json
```diff
@@ -123,9 +123,10 @@
   "scripts": {
     "vscode:prepublish": "pnpm compile",
     "release": "bumpp package.json --commit --push --tag && git update-ref refs/heads/release refs/heads/main && git push origin release",
-    "compile": "tsup ./src/extension.ts --external vscode --minify",
-    "watch": "tsup ./src/extension.ts --external vscode --watch --sourcemap",
-    "test": "vitest run --config vite.global-config.ts",
+    "compile": "tsup ./src/extension.ts test/**.test.ts --external vscode --minify",
```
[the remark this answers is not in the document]
sheremet-va: Ah I don’t think vitest exposes a similar api, but maybe we can have the same script, but run it with “tsx script” instead of “node script” so it supports ts?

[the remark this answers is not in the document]
hi-ogawa: Vscode's integration test is not so modern and it seems tricky to use typescript directly.

What I ended doing was to  create a wrapper to do `require("tsx/cjs")`  then re-export a custom "test runner script" https://github.com/hi-ogawa/vscode-extension-shell-shortcut/blob/main/src/test/run-wrapper.js

samples/basic/test/add.test.ts
```diff
@@ -1,17 +1,17 @@
-import { describe, expect, it } from 'vitest'
+import { expect } from 'chai'
```
hi-ogawa, line 1: These are Vitest sample projects, so why changing this too?

test/pure/testName.test.ts
hi-ogawa, line 1: I think we can still keep Vitest based unit test for these small utilities?

hi-ogawa:
Thanks for getting this started! I guess this PR is essentially for this?
- https://github.com/vitest-dev/vscode/issues/249

package.json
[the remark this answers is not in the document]
ffMathy: Can't we just make it easy to start `pnpm watch:test`? That would make everything behave nicely. The compilation time is almost instant.

`vscode-test` also has a watch mode. We could utilize that too.

And if we don't like building tests, perhaps it'd feel better if we had a separate command that would only build the tests, and another command to only build the source?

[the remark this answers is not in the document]
hi-ogawa: I don't necessary oppose transpiling vscode test files, but if there's a way to just plug-in `tsx` somewhere while still using `vscode-test` cli, then that would be great.

I haven't checked `vscode-test` cli, so let me dig into it a bit.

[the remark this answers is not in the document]
hi-ogawa: I think we can use `mocha.preload` option to setup just `require("tsx/cjs")`. I tested it here https://github.com/hi-ogawa/vscode-extension-shell-shortcut/pull/11

samples/basic/test/add.test.ts
ffMathy: Ah yes. That's a brain fart moment and an oversight. Will correct it soon.

test/pure/testName.test.ts
ffMathy: If the Vscode instance for testing is already running anyway, there is no need performance-wise. 

I feel like it'll confuse the reader of the code on what to pick at what situation, and I feel like using two test frameworks in one is less efficient and doesn't feel right. 

Let me know your thoughts on this.

package.json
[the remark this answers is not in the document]
ffMathy: Great! Whenever I have time, I'll take a look at that! 😍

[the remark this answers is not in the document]
ffMathy: That seemed to work! Awesome ❤️

ffMathy:
Review comments have been solved.

package.json
```diff
@@ -26,7 +26,7 @@
   "categories": [
     "Testing"
   ],
-  "main": "./dist/extension.js",
+  "main": "./dist/src/extension.js",
```
sheremet-va, line 29: It shouldn't be bundled in src anymore

.vscode-test.mjs
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from '@vscode/test-cli'
+
+export default defineConfig({
+  files: 'test/**/*.test.ts',
+  mocha: {
+    ui: 'bdd',
+    preload: './mocka.preload.cjs',
```
sheremet-va, line 7: Can't we just have `tsx/cjs` here?

package.json
```diff
@@ -125,7 +125,8 @@
     "release": "bumpp package.json --commit --push --tag && git update-ref refs/heads/release refs/heads/main && git push origin release",
     "compile": "tsup ./src/extension.ts --external vscode --minify",
     "watch": "tsup ./src/extension.ts --external vscode --watch --sourcemap",
-    "test": "vitest run --config vite.global-config.ts",
+    "test": "vscode-test",
+    "test:watch": "vscode-test --watch-files src/**/*.ts --watch-files test/**/*.test.tsx --watch-files test/**/*.test.tsx",
```
sheremet-va, line 129: 1. Do we have any tsx test files?
2. Why is the second pattern repeated?

samples/basic/test/using.test.ts
```diff
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest';
```
sheremet-va, line 1: Can you just revert all changes to samples?

ffMathy: Yeah sure. Good idea! 😊

package.json
ffMathy: 1. Good question 😏 
2. Good question 😏

package.json
ffMathy: Good point!

ffMathy:
@sheremet-va I fixed your review comments as well now.

sheremet-va (approved):

test/runner.test.ts
```diff
@@ -1,82 +1,44 @@
...
-    expect(output).toBe(`vitest.cmd=npx vitest.args=${expectedArgs} workspace=/test customStartProcess=${useCustomStartProcess}`)
+  const testCases = [
+    {isWindows: false, useCustomStartProcess: false, expectedArgs: `vitest,abc.spec.ts,-t,a (b) "c" d`},
+    {isWindows: false, useCustomStartProcess: true, expectedArgs: `vitest,abc.spec.ts,-t,a (b) "c" d`},
+    {isWindows: true, useCustomStartProcess: false, expectedArgs: `vitest,abc.spec.ts,-t,"a (b) \\"c\\" d"`},
+    {isWindows: true, useCustomStartProcess: true, expectedArgs: `vitest,abc.spec.ts,-t,a (b) "c" d`},
+  ]
+  testCases.forEach(function (testCase) {
+    const {isWindows, useCustomStartProcess, expectedArgs} = testCase
+    describe(
```
hi-ogawa, line 20: I think these tests is not running because of `describe` instead of `it`. After replace it with `it`, the tests fail with:

```
  1) TestRunner
       scheduleRun wrap test patterns if needed, (isWindows: false, customStartProcess: false):
     TypeError: Cannot redefine property: isWindows
        at Function.defineProperty (<anonymous>)
        at Context.<anonymous> (test/runner.test.ts:2:1709)
        at process.processImmediate (node:internal/timers:476:21)

  2) TestRunner
       "after each" hook for "scheduleRun wrap test patterns if needed, (isWindows: false, customStartProcess: false)":
     TypeError: Cannot redefine property: isWindows
        at Function.defineProperty (<anonymous>)
        at Context.<anonymous> (test/runner.test.ts:2:989)
        at process.processImmediate (node:internal/timers:476:21)
```

Maybe, we actually need to mock something after all?

ffMathy: Interesting. Although that failure doesn't seem to happen due to lack of mocking.

hi-ogawa: I didn't realize at that time, but I think this test's purpose was not about "running" but only about cli argument sanitization as written on the left side of the diff:

> Mock runVitestWithApi, causing it to return its arguments as its output to allow us to assert their values

Probably the original author wanted to unit test code around here:

https://github.com/vitest-dev/vscode/blob/85aa7edfce24914bee6a849e61d935e8f0b97bb2/src/pure/runner.ts#L62-L96

PR #263, opened 2024-02-23
fix: Use precise regexp pattern for vitest run args

keroxp:
Currently we passes full test pattern to vitest run as regex by `-t` option. However it isn't precise pattern because it lacks `^` and `$` for exact matching.

```js
(1) test('testing', () => {})
(2) test('testing 1', () => {})
```

In above case, (2) also runs when (1) runs because `-t 'testing'` matches both. We need to pass `-t '^ testing$'` for test case and `-t '^ testing'` for suite.


ffMathy (approved):
This looks great! Thank you! ❤️ 💪

src/TestData.ts
```diff
@@ -112,17 +113,36 @@ export class TestFile {
...
+      parents.push(iter)
+      iter = iter.parent
+    }
+
+    parents.reverse()
+    // vitest's test task name starts with ' ' of root suite
+    if (parents.length)
+      return ` ${parents.reduce((a, b) => `${a + b.nameResolver.pattern} `, '')}${this.start.nameResolver.pattern}`
+    else
+      return ` ${this.start.nameResolver.pattern}`
```
hi-ogawa, line 146: This mysterious leading space is considered a bug, but currently Vitest is keeping it to not break intellij:
- https://github.com/vitest-dev/vitest/issues/4045
- https://github.com/vitest-dev/vitest/issues/4103

To suppose both now and when it gets removed, maybe using `\s?` here would work?

ffMathy: Good idea.

sheremet-va (approved):

PR #269, opened 2024-02-25
fix: reference VS Code settings when determining vitest version

cameronbrill:
I am using yarn 4 pnp in my project, which is causing the default vitest binary resolution strategy (i.e. digging through `node_modules`) to fail. Given that we can pass in a command with which to run tests, I believe we should be able to use that command to determine the version of vitest. 

Look at the commit history for a summary of each change. I debugged this change by referencing [this PR](https://github.com/cameronbrill/portfolio/pull/20) in the `.vscode/launch.json` debug configuration. After these changes, I can finally run vitest tests in the native VS Code ui!

I appreciate any reviews!


sheremet-va (dismissed):

sheremet-va:
Can you fix linting errors?

sheremet-va (approved):

Issue #136, opened 2023-02-23
Coverage Reporting


gavinroderick:
**Is your feature request related to a problem? Please describe.**
I run unit tests through the extension, and would like to also run coverage reports in it somehow.  At the moment, I run `vitest run --coverage` in the terminal

**Describe the solution you'd like**
At first a button that will run the above terminal command for me. 

Long term, I'd like to be able to show a panel/tab that has lines covered/uncovered, similar to other IDE's (Jetbrains example picture below). 

**Describe alternatives you've considered**
None I can think of, this is just a nice to have really. Happy to continue with my current method if you want to keep the extension purely unit test based. 

**Additional context**
I'm new to open source, but would be keen to give an implementation of this a shot! If you think it's worthwhile, I can look at submitting a PR. 

In the first instance, I'd probably just try to have a button that checks to see if a coverage reporter is installed, and then run it in the terminal. If that goes well, then maybe look at getting fancier with displaying the output. 

![Jetbrains Coverage Reporter]()

thanks, 

Gavin

MDUK0001:
Jest VS Code extension has this - if you're thinking of starting this it may be worth looking at their implementation for inspiration: https://github.com/jest-community/vscode-jest

adueck:
Yes +1 for this. I love using this [Jest VS Code extension](https://github.com/jest-community/vscode-jest). With the `toggle coverage` command you can see the coverage show up with lines red and yellow (or whatever color you set) over the uncovered and partially covered lines in the file. It's great!

simonknittel:
I'm able to achieve this using this extension: https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters

You'll have to add `"vitest.commandLine": "npx vitest --coverage"` to your workspace settings and configure Vitest to use the Instabul with the lcov reporter.

**.vscode/extensions.json**
```extensions.json
{
  "recommendations": [
    "zixuanchen.vitest-explorer",
    "ryanluker.vscode-coverage-gutters"
  ]
}
```

**.vscode/settings.json**
```settings.json
{
  "vitest.enable": true,
  "vitest.commandLine": "npx vitest --coverage"
}
```

**vitest.config.ts**
```vitest.config.ts
import { defineConfig } from "vitest/config";

export default defineConfig({
  test: {
    coverage: {
      provider: "istanbul",
      reporter: [
        "text", // For the terminal
        "lcov", // For the VSCode extension and browser
      ],
    },
  },
});
[...]

PR #253, opened 2024-02-19
feat: use Vitest public API instead of Vitest CLI

sheremet-va:
This PR switches to use public Vitest API instead of calling `vitest` CLI directly. Improvements that come with it:

- Requires Vitest 1.4.0 or higher
- Faster test execution (instant feedback when clicking on "start")
- Better continues mode integration
- Support for "each" tests
- Support in-source tests
 
Known issues
- Currently, the extension requires `vite.config` or `vitest.config` file to function
- The "continues" mode doesn't update test results if several test functions are selected - it will use the last one selected. For now, you can use continues mode on a file without problems
- Sometimes debugger starts only on a second try
- When using debugger, tests are not marked as "enqueued"

PR #275, opened 2024-03-02
Add basic coverage code lens support [issue-136]

uioporqwerty:
image

Adds a code lens similar to jest as documented here: https://github.com/vitest-dev/vscode/issues/136#issuecomment-1514106636

Current assumptions:

- User has `json` as their reporter.
- Customer coverage reporters unsupported for now.

Improvements and gutter support to follow, but this is a good first step I think. Feedback welcome :)


src/runHandler.ts
```diff
@@ -76,6 +76,8 @@ export async function runHandler(
 
   run.end()
   log.info('Tests run end')
+
+  await vscode.commands.executeCommand('editor.action.refreshCodeLens');
```
uioporqwerty, line 80: We will refresh all the code lens here after the test so ensure they have up to date coverage information

test/coverageCodeLensProvider.test.ts
```diff
@@ -0,0 +1,92 @@
+import * as vscode from 'vscode';
+import * as fs from 'fs'
+import { describe, expect, it, vi, beforeEach, afterEach } from 'vitest'
```
hi-ogawa, line 3: Tests are now executed by https://github.com/microsoft/vscode-test-cli with mocha, so I didn't think this would work. Did you run `pnpm test`?

sheremet-va:

Feel free to reimplement this feature in a separate PR. It should be a bit easier because we have direct access to the coverage, and we don't need to read it from the reporter file.

Issue #259, opened 2024-02-23
Regression: Duplicate describes can't be run


ffMathy:
**Describe the bug**
After merging #214, a bug has been introduced for the following scenario where there are duplicate describes:

```typescript
import { describe, expect, it } from 'vitest'
import { add } from '../src/add'

describe('duplicate describe', () => {
  it('test 1', () => {
    expect(add(1, 1)).toBe(2)
  })
})

describe('duplicate describe', () => {
  it('test 2', () => {
    expect(add(1, 1)).toBe(2)
  })
})
```

In the above scenario, you can't run "test 2". It will display a "test not found" error. This was not the case before.

**To Reproduce**
Add the above test and try to run the "test 2" test.

**Expected behavior**
I expected the test to be run.

**Environment**
N/A

**Additional context**
N/A

keroxp:
This is introduced indeed by #214 but I guess it didn't work in previous version. When one case ran from vscode UI, actually both cases ran and wrong statuses were synced. This is because there're no way to distinguish those two blocks based on vitest's test results. UI is built based on the position in the file, but results have no information except their name.

ffMathy:
Are you saying this is fixed by #263?

keroxp:
No, I actually started it for this, but those patch can't resolve this. Issue still exists.

ffMathy:
Ah okay. In my opinion, it's fine if both tests run. That's how it was before, and I guess also how it works when filtering with vitest directly.

sheremet-va:
This should be fixed in pre-release 0.5.0 and higher. Note that the extension now requires Vitest 1.4.0 or higher.

Currently it will run _both_ described. We can still do better - I wanted to introduce a feature to run tests in this form in Vitest itself (maybe in 1.5.0):

```ts
vitest test-file.test.js:13
```

PR #283, opened 2024-03-08
feat: show basic diagnostics via LSP

keroxp:
Implemented basic diagnostic informations for test codes via VSCode Language Server. 
https://code.visualstudio.com/api/language-extensions/language-server-extension-guide

Now, warnings will be shown on undesirable test cases:
image

Related to #259, duplicated desc/test blocks in same hierarchy can't be ran now. This feature aims to notice users undesirable naming cases. There're 3 diagnostic informations:

- Duplicated name blocks in same hierarchy 
- Empty name block
- `.each` blocks that has no placeholder

I'm novice of LSP. Would someone give me an advise?

### DIFF
--- a/package.json
+++ b/package.json
@@ -169,6 +169,9 @@
     "tsx": "^4.7.1",
     "typescript": "^5.3.3",
     "vitest": "^1.3.1",
+    "vscode-languageclient": "^9.0.1",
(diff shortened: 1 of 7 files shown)


keroxp:
I close PR as they will be resolved when #253 is merged.

PR #284, opened 2024-03-10
feat(vitest): add yarn3 + pnp support

### LINKED ISSUE #285: No vitest config detected on yarn + pnp + nx projects
### Describe the bug

No tests are detected on Yarn + PnP projects. The logs show the following lines

```
Vitest Workspace [workspace]: Vitest version = undefined
ERR [Extension Host] Cannot run tests: no Vitest config found.
```

### Reproduction

I cannot share the project code I used to verify it, but I believe the investigation below might be enough to highlight the problem.

The function https://github.com/vitest-dev/vscode/blob/main/src/pure/utils.ts#L38 is triggered at https://github.com/vitest-dev/vscode/blob/4283697a5d638105dfabf83ec417f55c5f0669cd/src/config.ts#L81 and only checks for the script at node_modules folder. Yarn PnP projects do not have such folder and, therefore,the resulting command is null. As a consequence, the vitest version is undefined at https://github.com/vitest-dev/vscode/blob/4283697a5d638105dfabf83ec417f55c5f0669cd/src/config.ts#L84

The proposed solutions:
- Check for yarn.lock and require a script "yarn vitest".
- Detect for .pnp files. If they exist, consider the command as "yarn vitest".

Additionally, there is an issue with the execution of `chunksToLinesAsync(child.stdout)` at https://github.com/vitest-dev/vscode/blob/4283697a5d638105dfabf83ec417f55c5f0669cd/src/pure/utils.ts#L113 which empty when the command is correct on the system informed in this bug report. This will result in a failure when loading the version through yarn. 

### System Info

```shell
GNU/Linux, VSCode.
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vscode/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vscode/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).

saulotoledo:
Closes #285.

### DIFF
--- a/src/pure/utils.ts
+++ b/src/pure/utils.ts
@@ -50,10 +50,19 @@ export function getVitestCommand(
     }
   }
 
-  const node_modules = path.resolve(projectRoot, 'node_modules')
   try {
-    if (!existsSync(node_modules))
+    const yarnLock = path.resolve(projectRoot, 'yarn.lock')
+    if (existsSync(yarnLock)) {
+      return {
+        cmd: 'yarn',
+        args: ['vitest'],
+      }
+    }
+
+    const node_modules = path.resolve(projectRoot, 'node_modules')
+    if (!existsSync(node_modules)) {
       return getVitestCommand(path.dirname(projectRoot))
+    }
 
     const suffixes = ['']
     if (isWindows)
@@ -106,19 +115,26 @@ export async function spawnVitestVersion(command: string, args: string[], env?: 
     env,
   })
 
-  child.on('error', () => {
-    log.info('Command not found')
-  })
+  return new Promise((resolve) => {
+    child.stdout.on('data', (data) => {
+      child.kill()
+      log.info(data.toString())
 
-  for await (const line of chunksToLinesAsync(child.stdout)) {
-    child.kill()
-    log.info(line)
+      const match = data.toString().match(/vitest\/(\d+.\d+.\d+)/)
 
-    const match = line.match(/vitest\/(\d+.\d+.\d+)/)
+      if (match && match.length > 0)
+        resolve(match[1])
+    });
 
-    if (match && match.length > 0)
-      return match[1]
-  }
+    child.on('error', () => {
+      log.info('Command not found')
+      resolve(undefined);
+    })
+
+    child.on('close', (code) => {
+      resolve(undefined);
+    });
+  });
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/pure/utils.ts
```diff
@@ -50,10 +50,19 @@ export function getVitestCommand(
     }
   }
 
-  const node_modules = path.resolve(projectRoot, 'node_modules')
   try {
-    if (!existsSync(node_modules))
+    const yarnLock = path.resolve(projectRoot, 'yarn.lock')
+    if (existsSync(yarnLock)) {
```
saulotoledo, line 55: To keep it simple, I just detect Yarn here and return "yarn vitest" right away. This might have side effects for projects running yarn and using node_modules and have a "vitest" script in their package.json. In a first thought, I think this might be the desired result anyways, so we do not need to find for the binary manually since we can do it through the package manager. However, if this is a problem, an alternative solution is to replace the two lines above with the solution below, which will only check for .pnp files. Thus, the projects with yarn and node_modules will fallback to the original implementation:

```
const hasPnpFiles = readdirSync(projectRoot).filter(file => /\.pnp\.(?:c|m)?js$/.test(file)).length > 0
if (hasPnpFiles) {
```

For this to work, we also need to import `readdirSync` at `import { existsSync, readdirSync } from 'fs-extra'`.

Please let me know if you would prefer this alternative.

src/pure/utils.ts
```diff
@@ -106,19 +115,26 @@ export async function spawnVitestVersion(command: string, args: string[], env?:
...
-  }
+    child.on('error', () => {
+      log.info('Command not found')
+      resolve(undefined);
+    })
+
+    child.on('close', (code) => {
+      resolve(undefined);
+    });
+  });
```
saulotoledo, line 137: For a reason I did not investigate further, `chunksToLinesAsync(child.stdout)` returns empty here, and the version detection fails. I replaced the lib call with a plain promise. This implementation should work on every system. I am also resolving everything to `undefined` in the other cases because of the behavior described on the tests.

saulotoledo:
I added highlights above explaining the decision taken. Please let me know if you require improvements or other changes in this proposal.

sheremet-va:

It doesn't support pnp yet, feel free to open a separate PR to fix that.

PR #316, opened 2024-03-21
fix: yarn3+ pnp support

### LINKED ISSUE #285: No vitest config detected on yarn + pnp + nx projects

saulotoledo:
Closes #285.


src/api/resolve.ts
```diff
@@ -38,6 +39,8 @@ export function resolveVitestPackage(cwd: string, folder: vscode.WorkspaceFolder
 
 export function resolveVitestPackagePath(cwd: string, folder: vscode.WorkspaceFolder) {
   const customPackagePath = getConfig(folder).packagePath
+  if (existsSync(`${cwd}/.pnp.cjs`))
+    return null
```
saulotoledo, line 43: This will allow the extension to work when we have a yarn + pnp project inside a project running node (like the sample added to this project). If this is not done, we will fallback to `require.resolve()` below and the yarn project will be ignored and the extension does not work.

src/api.ts
```diff
@@ -188,13 +190,15 @@ function resolveVitestConfig(showWarning: boolean, configOrWorkspaceFile: vscode
   }
 
   if (vitest.pnp) {
-    // TODO: try to load vitest package version from pnp
+    const zipOpenFs = new ZipOpenFS({ libzip: getLibzipSync() })
+    const crossPlatformCompatFs = new PosixFS(zipOpenFs)
+    const version = JSON.parse(crossPlatformCompatFs.readFileSync(vitest.vitestPackageJsonPath, 'utf8')).version
```
saulotoledo, line 195: We cannot rely on the versions in the reference attribute from [packagelocator](https://yarnpkg.com/advanced/pnpapi#packagelocator) because the documentation states: "reference field should be considered an opaque string whose value may be **whatever** the PnP implementation decides to put there."

Thus, I followed the suggestion from [this link](https://yarnpkg.com/advanced/pnpapi#accessing-the-files) in the Yarn documentation to get the version directly from the library package.json.

test-e2e/basic.test.ts
```diff
@@ -52,3 +52,25 @@ test('imba', async ({ launch }) => {
   await expect(page.locator(`[title*="utils.imba (Passed)"]`)).toBeVisible()
   await expect(page.locator(`[title*="counter.imba (Failed)"]`)).toBeVisible()
 })
+
+test.only('yarn-pnp', async ({ launch }) => {
```
saulotoledo, line 56: To be updated later.

samples/yarn-pnp/.yarn/releases/yarn-4.1.1.cjs
saulotoledo, line 1: Let me know if you prefer to remove this file from here.

src/api.ts
sheremet-va: This is an insane amount of code that we need to load for everyone just to do this. Maybe instead we can do something like this: https://github.com/microsoft/playwright-vscode/blob/main/src/playwrightFinder.ts

saulotoledo: I could, in theory, get it directly from the project's `package.json`, but the version might not be exact (e.g., `>=2.0.0`), and there are cases such as version ranges and similar that will make it harder for us. I will check the suggested resolver to see if I can extract it from Vitest.

saulotoledo: I used `yarn exec` here as well since we are going to use it for the test execution, as discussed on the issue thread.

hi-ogawa:
>  It fails because yarn requires dependencies https://github.com/vitest-dev/vscode/issues/285#issuecomment-2013624894

I'm just testing this PR locally and what exactly is the error you're seeing currently?

It looks like there's an error when trying to import `vitest/node` via full path like this, but is Yarn pnp meant to work in that way? For example, running it directly with `yarn node` doesn't seem to work either.

```ts
// repro.mjs
await import("file:///home/hiroshi/.yarn/berry/cache/vitest-npm-1.4.0-465b7cb84c-10c0.zip/node_modules/vitest/dist/node.js")
```

```sh
$ yarn node repro.mjs 

node:internal/process/esm_loader:34
      internalBinding('errors').triggerUncaughtException(
                                ^
Error: vitest tried to access pathe, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: pathe (via "pathe/package.json")
Required by: vitest@npm:1.4.0 (via /home/hiroshi/.yarn/berry/cache/vitest-npm-1.4.0-465b7cb84c-10c0.zip/node_modules/vitest/dist/node.js)

    at makeError (/home/hiroshi/code/others/vitest-vscode/samples/yarn-pnp/.pnp.cjs:7353:34)
    at resolveToUnqualified (/home/hiroshi/code/others/vitest-vscode/samples/yarn-pnp/.pnp.cjs:9001:21)
    at Object.resolveToUnqualified (/home/hiroshi/code/others/vitest-vscode/samples/yarn-pnp/.pnp.cjs:9181:26)
    at resolve$1 (file:///home/hiroshi/code/others/vitest-vscode/samples/yarn-pnp/.pnp.loader.mjs:1993:31)
    at nextResolve (node:internal/modules/esm/hooks:865:28)
    at Hooks.resolve (node:internal/modules/esm/hooks:303:30)
    at handleMessage (node:internal/modules/esm/worker:196:24)
    at checkForMessages (node:internal/modules/esm/worker:138:28)
    at process.<anonymous> (node:internal/modules/esm/worker:157:5)
    at process.emit (node:events:518:28)
```

So, I'm wondering if we can just use `import("vitest/node")` and then yarn pnp loader will figure out everything without resolving path on our own?

samples/yarn-pnp/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@vitest/vscode-sample-yarn-pnp",
```
hi-ogawa, line 2: Corepack is confused locally, so probably we need to add `packageManager` explicitly here:

```diff
diff --git a/samples/yarn-pnp/package.json b/samples/yarn-pnp/package.json
index a1ee270..68c9864 100644
--- a/samples/yarn-pnp/package.json
+++ b/samples/yarn-pnp/package.json
@@ -3,6 +3,7 @@
   "type": "module",
   "private": true,
   "version": "1.0.0",
+  "packageManager": "yarn@4.1.1",
   "scripts": {
     "test": "vitest"
   },
```

saulotoledo: Agreed. Will do.

sheremet-va:
> Also, loader setup here is not necessary since --require and --experimental-loader are already setup via execArgv?

This option is deprecated in the latest Node.js version:

```
--experimental-loader` may be removed in the future; instead use `register()`
```

saulotoledo:
@sheremet-va Should we push the first version as of now and create another PR for removing the deprecated option? The proposal by @hi-ogawa just worked. I am investigating one issue with workspaces where vitest is not detected for yarn. I can push it here as soon as I have fixed it.

sheremet-va:
> @sheremet-va Should we push the first version as of now and create another PR for removing the deprecated option?

I don't see a reason why we shouldn't do it here

saulotoledo:
@sheremet-va ok. I am checking the extension against another project I have, and it is not working. I will debug that first, and then I will get back to this case. One thing I just found is that `gte()` from `semver` is always returning `true` after minified, so it is not detecting versions below the minimum. For now I just updated the versions of the project to investigate my current issue, but it would be nice to check that issue as well. I checked it on GNU/Linux.

saulotoledo:
Hi @hi-ogawa and @sheremet-va. I did not find time to work on this in the past few weeks, but I started looking into it today. Before I continue, I would like to check whether my understanding is correct below. Please also let me know if you have better suggestions.

From now on, the main idea is to use `yarn` directly instead of the Vitest API. I do not know the entire code of the extension, but my first idea after looking into the code is to spawn the yarn process at the `worker.ts` file. I would create two versions of the `initVitest` method, one using the Vitest API and one running `yarn`. My idea was to create a wrapper and keep the rest of the code untouched.

However, I noticed that the extension is somewhat "coupled" to the Vitest instance. For example, `VSCodeReporter`, `createWorkerRPC` and their dependencies depend directly on instances of `Vitest` (from the Vitest API). I would like to avoid creating an entirely new structure to run Yarn. It would be better to have a standard interface between those, or we risk missing functionality and having to support two different implementations.

That said, do you have any suggestions on how I should proceed? I welcome any ideas so I can provide a solution that is aligned with your project plans. Thanks in advance!

sheremet-va:
The idea is not to use yarn to run tests, the idea is to use yarn exec to spawn the worker.js file, then the only thing you need to figure out is communication with that file.

Issue #319, opened 2024-03-23
Unable to resolve tests in monorepo with vitest.workspace.ts + root vitest.config.ts


chriswheeldon-peakon:
### Describe the bug

The plugin resolves no tests in a monorepo with a vitest.workspace.ts and root vitest.config.ts. I do not see any errors in the plugin logs.

I am using pre-release version 0.5.9 of the plugin.

### Reproduction

The quickest way to reproduce is to add an empty or minimal (`export default defineConfig({});`) vitest.config.ts to the root of the monorepo-vitest-workspace sample ([here](https://github.com/vitest-dev/vscode/tree/main/samples/monorepo-vitest-workspace)).

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 2.06 GB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - /var/folders/sb/jdkc4cv91zvgvvzwhq9lyc6w0000gp/T/fnm_multishells/37003_1711106402798/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.4 - /var/folders/sb/jdkc4cv91zvgvvzwhq9lyc6w0000gp/T/fnm_multishells/37003_1711106402798/bin/npm
    pnpm: 8.15.3 - /var/folders/sb/jdkc4cv91zvgvvzwhq9lyc6w0000gp/T/fnm_multishells/37003_1711106402798/bin/pnpm
  IDEs:
    VSCode: 1.87.2 - /Applications/Visual Studio Code.app/Contents/Resources/app/bin/code
    Vim: 9.0 - /usr/bin/vim
    Xcode: 15.3/15E204a - /usr/bin/xcodebuild
  Browsers:
    Chrome: 123.0.6312.59
    Safari: 17.3.1
  npmPackages:
    vitest: ^1.4.0 => 1.4.0
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitest-dev/vscode/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitest-dev/vscode/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitest.dev/guide/).
[...]

PR #322, opened 2024-03-25
fix: fix workspace with root `vite.config.ts`

hi-ogawa:
- closes https://github.com/vitest-dev/vscode/issues/319

Extension side is requesting `rpc.getFiles(".../vitest.workspace.ts")`, but worker side had `vitestById = { ".../vitest.config.ts": Vitest }` due to `getId` picking up root config instead of workspace config. 

This was leading to a following `undefined` error since `vitestById[".../vitest.workspace.ts"] === undefined`:

```
2024-03-25 13:38:19.276 [error] TypeError: Cannot read properties of undefined (reading 'server')
    at getId (/home/hiroshi/code/others/vitest-vscode/dist/worker.js:925:20)
    at globTestFiles (/home/hiroshi/code/others/vitest-vscode/dist/worker.js:949:27)
    at Proxy.getFiles (/home/hiroshi/code/others/vitest-vscode/dist/worker.js:995:27)
```

## test plan

I manually tested on `samples/monorepo-vitest-workspace`.

### DIFF
--- a/src/worker/actions.ts
+++ b/src/worker/actions.ts
@@ -59,7 +59,7 @@ export function createWorkerMethods(vitest: Vitest[]): BirpcMethods {
   const vitestEntries = Object.entries(vitestById)
 
   function getId(vitest: Vitest) {
-    return vitest.server.config.configFile || vitest.config.workspace || vitest.config.root
+    return vitest.config.workspace || vitest.server.config.configFile || vitest.config.root
   }
 
   async function rerunTests(vitest: Vitest, files: string[]) {
--- a/src/worker/actions.ts
+++ b/src/worker/actions.ts
@@ -4,12 +4,8 @@ import type { BirpcMethods } from '../api/rpc'
 
 const _require = require
 
-export function createWorkerMethods(vitest: Vitest[]): BirpcMethods {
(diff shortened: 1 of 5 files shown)


src/worker/actions.ts
```diff
@@ -59,7 +59,7 @@ export function createWorkerMethods(vitest: Vitest[]): BirpcMethods {
   const vitestEntries = Object.entries(vitestById)
 
   function getId(vitest: Vitest) {
-    return vitest.server.config.configFile || vitest.config.workspace || vitest.config.root
+    return vitest.config.workspace || vitest.server.config.configFile || vitest.config.root
```
sheremet-va, line 62: I was thinking about injecting ID somehow - we know it in the `worker.ts` file

hi-ogawa: I think I know what you mean. Yeah, probably `getId(vitest)` can be cleaned up since we have `initVitest(meta)` is already using `meta.id`.

I'll see what I can do.

hi-ogawa: I removed `getId` in favor of constructing `vitestById` ahead of time https://github.com/vitest-dev/vscode/pull/322/commits/91f30446fe73d3bcb99e9e5e23b1b1a856cf34b8

sheremet-va (approved):

PR #328, opened 2024-04-02
Enhance issue reporting template with version validation

MilanKovacic:
This PR introduces a validation step in the issue reporting template to ensure that issues are reported against the latest version of the extension. The rationale behind this change is to streamline the issue resolution process by focusing efforts on current problems, reducing the time spent on issues already addressed in recent updates, and encouraging users to keep their installations updated.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -43,6 +43,8 @@ body:
       label: Validations
       description: Before submitting the issue, please make sure you do the following
       options:
+        - label: Check that you are using the latest version of the extension
+          required: true
         - label: Check that there isn't [already an issue](https://github.com/vitest-dev/vscode/issues) that reports the same bug to avoid creating a duplicate.
           required: true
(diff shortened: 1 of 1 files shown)


sheremet-va:
Do we need the version input now?

ffMathy:
The version output, doesn't that only include the Vitest packages? This would be for the version of the extension itself.

ffMathy:
The version output, doesn't that only include the Vitest packages? This would be for the version of the extension itself.

sheremet-va:
> The version output, doesn't that only include the Vitest packages? This would be for the version of the extension itself.

Please, look at the current issue template. It has "Output" (logs from the "Output" window) and "Version" fields. If we include this check, then why do we need the "Version" field?

MilanKovacic:
The primary reason for keeping the version field, even with the required version validation step, is to ensure we can easily see the specific extension version that the issue was reported against.

Over time, the "latest version" information from the validation step may become less meaningful. The version field provides a clear, accessible record of the exact version used when the issue was reported.

PR #309, opened 2024-03-20
fix: set env vars `TEST`, `VITEST`, and `NODE_ENV`

hi-ogawa:
- closes https://github.com/vitest-dev/vscode/issues/305

PR #346, opened 2024-04-15
Default `NODE_ENV` to `test`

john-griffin:
Followup to https://github.com/vitest-dev/vscode/pull/309

When running the extension on a fresh install I could see the `NODE_ENV` by default is coming through as `true`. If we want to match the behavior of `startVitest` in  https://github.com/vitest-dev/vitest/blob/5c7e9ca05491aeda225ce4616f06eefcd068c0b4/packages/vitest/src/node/cli/cli-api.ts#L40 we should default to `test`.

### DIFF
--- a/samples/basic/test/env.test.ts
+++ b/samples/basic/test/env.test.ts
@@ -3,7 +3,7 @@ import { test, expect } from "vitest";
 test('process.env', () => {
   expect(process.env.TEST).toBe('true');
   expect(process.env.VITEST).toBe('true');
-  expect(process.env.NODE_ENV).toBe('true');
+  expect(process.env.NODE_ENV).toEqual('test');
   expect(process.env.VITEST_VSCODE).toBe('true');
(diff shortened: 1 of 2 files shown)


samples/basic/test/env.test.ts
```diff
@@ -3,7 +3,7 @@ import { test, expect } from "vitest";
 test('process.env', () => {
   expect(process.env.TEST).toBe('true');
   expect(process.env.VITEST).toBe('true');
-  expect(process.env.NODE_ENV).toBe('true');
+  expect(process.env.NODE_ENV).toEqual('test');
```
sheremet-va, line 6:
```suggestion
  expect(process.env.NODE_ENV).toBe('test');
```

sheremet-va (approved):

sheremet-va:
Thanks 👍🏻

hi-ogawa:
Omg, bad typo... Thanks for the fix!

PR #384, opened 2024-05-16
bugfix(utils): add volta support on win32

### LINKED ISSUE #382: Volta support / Process exited with code 0
### Describe the bug

When running visual studio code on a system using volta (With lts installed as the default node version). The visual studio code extension fail to start.

It seems to open the node process with the  `"C:\Program Files\Volta\node.EXE" --dns-result-order=ipv4first C:/Users/jbzeller/.vscode/extensions/vitest.explorer-0.10.2/dist/worker.js` command line. which seems to open then close immediately without error. The command is then relayed to the real node installation.

### Reproduction

1. Use https://github.com/vitest-dev/vitest/tree/main/examples/basic as a basic sample folder.
2. Install volta on your machine (uninstalling existing node installation first).
3. volta install node@20
4. In the folder run npm install
5. Open visual studio code with the vitest extension installed.

### Output

```shell
[INFO 16:14:41] [v0.10.2] Vitest extension is activated because Vitest is installed or there is a Vite/Vitest config file in the workspace.
[INFO 16:14:42] [API] Running Vitest: v1.6.0 (vite.config.ts)
[INFO 16:14:42] [API] Starting Vitest process with Node.js: C:\Program Files\Volta\node.EXE
[Error 16:14:43] [Error Error] Vitest process exited with code 0
Error: Vitest process exited with code 0
	at ChildProcess.d (c:\Users\jbzeller\.vscode\extensions\vitest.explorer-0.10.2\dist\extension.js:21:170)
	at ChildProcess.emit (node:events:517:28)
```

### Version

0.10.2

### Validations

- [X] Check that you are using the latest version of the extension

Esurnir:
This only works if the installation folder is "Volta".
I don't know how to check for volta otherwise.

I only have a windows machine, I don't know volta's behaviour on linux so I can't check what will happen there.

Fix #382

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -107,7 +107,7 @@ export async function findNode(cwd: string): Promise<string> {
 
   // If volta isn't installed in a Volta folder, this test will fail.
   // If anyone got a better idea for checking volta's presence, please let me know.
-  const voltaRegex = /Volta/i
+  const voltaRegex = /Volta\\node\.exe/i
 
   // Stage 4: We have found Node.js, but it might be managed by Volta.
   // This attempt to ask volta for the path to the node executable.
@@ -124,7 +124,7 @@ function findNodeUsingVoltaOnWindows(cwd: string): Promise<string | undefined> {
   if (process.platform !== 'win32')
     return Promise.resolve(undefined)
   return new Promise<string | undefined>((resolve) => {
-    const childProcess = spawn(`volta which node`, {
+    const childProcess = spawn('volta which node', {
       stdio: 'pipe',
       shell: true,
       cwd,
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -104,10 +104,40 @@ export async function findNode(cwd: string): Promise<string> {
   // Stage 3: If we still haven't found Node.js, try to find it via a subprocess.
   // This evaluates shell rc/profile files and makes nvm work.
   node ??= await findNodeViaShell(cwd)
+
+  // If volta isn't installed in a Volta folder, this test will fail.
+  // If anyone got a better idea for checking volta's presence, please let me know.
+  const voltaRegex = /Volta/i
+
+  // Stage 4: We have found Node.js, but it might be managed by Volta.
+  // This attempt to ask volta for the path to the node executable.
+  if (node && voltaRegex.test(node))
+    node = await findNodeUsingVoltaOnWindows(cwd) ?? node
+
   if (!node)
     throw new Error(`Unable to find 'node' executable.\nMake sure to have Node.js installed and available in your PATH.\nCurrent PATH: '${process.env.PATH}'.`)
   pathToNodeJS = node
   return node
+}
+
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -104,12 +104,42 @@ export async function findNode(cwd: string): Promise<string> {
   // Stage 3: If we still haven't found Node.js, try to find it via a subprocess.
   // This evaluates shell rc/profile files and makes nvm work.
   node ??= await findNodeViaShell(cwd)
+
+  // If volta isn't installed in a Volta folder, this test will fail.
+  // If anyone got a better idea for checking volta's presence, please let me know.
+  const voltaRegex = /Volta/i
```
sheremet-va, line 110: I would check that it ends with Volta/node.EXE since it's hardcoded by Volta already

src/utils.ts
```diff
@@ -104,12 +104,42 @@ export async function findNode(cwd: string): Promise<string> {
...
     throw new Error(`Unable to find 'node' executable.\nMake sure to have Node.js installed and available in your PATH.\nCurrent PATH: '${process.env.PATH}'.`)
   pathToNodeJS = node
   return node
 }
 
+function findNodeUsingVoltaOnWindows(cwd: string): Promise<string | undefined> {
+  if (process.platform !== 'win32')
+    return Promise.resolve(undefined)
+  return new Promise<string | undefined>((resolve) => {
+    const childProcess = spawn(`volta which node`, {
```
sheremet-va, line 127: instead of "volta", we should probably pass down the found executable, no? Or is it a different one?

sheremet-va (changes requested):

src/utils.ts
Esurnir: The executable we have is node, It's a different one.

src/utils.ts
Esurnir: I have updated to Volta\node.exe. Windows use backslash for path even in node. The test would fail otherwise.

Esurnir:
For your information here's a reply I got from Chuck Pierce from the Volta discord to my question about detecting's Volta presence on Windows :

"For detecting the presence of Volta on Windows I could see a couple of similar approaches:

1. The approach of finding the path and searching it for Volta actually may be the least intrusive, since it’s a straightforward string match. It’s not _technically_ the most robust, as a user could theoretically have changed the install directory to be something else (that doesn’t also include “Volta” somewhere), but I would wager that’s quite rare.
2. Don’t try to detect it a priori, run `volta which node` and see if it succeeds. If it does, Volta is there and you get the value you need, if it fails, then you can fall back."

sheremet-va (approved):

qlagraula:
Hello, is there any particular reason not to use the `volta which node` command to detect Volta? I installed Volta through Scoop, which make the detection failed with the regex. Unfortunately, it looks like I can’t customize the install folder when using Scoop.

I believe using the `volta which node` command could be an effective solution. I’d be happy to open a pull request if you don’t see any drawbacks.

PR #440, opened 2024-07-26
chore: upgrade to eslint9

btea:
Upgrade `eslint` and `@antfu/eslint-config`, and fix some eslint errors.


sheremet-va (dismissed):

btea:
It looks like the plugin packaged code uses `commonjs` syntax, maybe we shouldn't add the `"type": "module"` field.

sheremet-va:
> It looks like the plugin packaged code uses `commonjs` syntax, maybe we shouldn't add the `"type": "module"` field.

Yeah, vscode extensions cannot be ESM

btea:
The test seems to be consistently failing, but it doesn't seem to be caused by my changes.

sheremet-va (approved):

PR #479, opened 2024-09-30
fix: handle config de-sync when using workspaces

### LINKED ISSUE #477: Coverage Reporting not displaying when using workspace
### Describe the bug

I have a few repos with a similar workspace/monorepo layout, I haven't been able to get the inline coverage reporting from VSCode working in any of them. When I run the test suite, the coverage panel does not appear and no coverage UI appears in the editor. The Test Results panel simply shows ` *  The test run did not record any output. ` and I also see `Coverage was disabled due to all requests being exhausted` in the output (see below).

The tests themselves do run and they show as passing, it's just the coverage that isn't working. Also running coverage via a script (in the repro case with `pnpm test:coverage`) works fine.

If I remove the workspace file and update the settings to point to an individual package config then coverage reporting does work. See the `without-workspace` branch in the reproduction to show this.

I feel like there is something misconfigured but having read the documentation thoroughly, both for Vitest and the extension, workspaces are stated to "just work". This configuration pattern does work fine when running tests locally and in CI so I would expect the extension to also support it.

### Reproduction

https://github.com/BenShelton/vitest-vscode-repro

### Output

```shell
[INFO 12:03:17] [v1.2.10] Vitest extension is activated because Vitest is installed or there is a Vite/Vitest config file in the workspace.
[INFO 12:03:17] [API] Using user workspace config: ***/vitest-vscode-repro/vitest.workspace.ts
[INFO 12:03:17] [API] Running Vitest v2.1.1 (vitest-vscode-repro/vitest.workspace.ts) with Node.js: ***/.nvm/versions/node/v20.15.1/bin/node 
[INFO 12:03:17] [API] Vitest v2.1.1 (vitest-vscode-repro/vitest.workspace.ts) process 9352 created
[INFO 12:03:28] [Worker] Running coverage with configuration: {
  provider: 'istanbul',
  enabled: true,
  all: true,
  clean: true,
  cleanOnRerun: true,
  reportsDirectory: '/var/folders/cz/xhwb1xys7v335hr7pzhssrx00000gn/T/vitest-coverage-cd8866ce-2443-4470-a3b1-0403d8dc0451',
  exclude: [
    'coverage/**',
    'dist/**',
    '**/node_modules/**',
    '**/[.]**',
    'packages/*/test?(s)/**',
    '**/*.d.ts',
    '**/virtual:*',
    '**/__x00__*',
    '**/\x00*',
    'cypress/**',
    'test?(s)/**',
    'test?(-*).?(c|m)[jt]s?(x)',
    '**/*{.,-}{test,spec,bench,benchmark}?(-d).?(c|m)[jt]s?(x)',
    '**/__tests__/**',
    '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build,eslint,prettier}.config.*',
    '**/vitest.{workspace,projects}.[jt]s?(on)',
    '**/.{eslint,mocha,prettier}rc.{?(c|m)js,yml}',
    'test/setup.ts'
  ],
  reportOnFailure: true,
  reporter: [ [ 'json', {} ] ],
  extension: [
    '.js',     '.cjs',
    '.mjs',    '.ts',
    '.mts',    '.tsx',
    '.jsx',    '.vue',
    '.svelte', '.marko'
  ],
  allowExternal: false,
  excludeAfterRemap: false,
  ignoreEmptyLines: true,
  processingConcurrency: 10,
  include: [ 'app/src/**/*.{ts,vue}' ]
}
[INFO 12:03:28] Running all tests in vitest-vscode-repro
[12:03:28] The runner is starting because tests ***/vitest-vscode-repro/app/src/__tests__/App.spec.ts ***/vitest-vscode-repro/app/src/utils/__tests__/time.spec.ts were started due to a file change
[12:03:28] Enqueuing "[App] App.spec.ts"
[12:03:28] Enqueuing "[App] time.spec.ts"
[12:03:28] No task result for "should format seconds as a timestamp", assuming the test just started running
[12:03:28] No task result for "should return the seconds between 2 times", assuming the test just started running
[12:03:28] No task result for "should resolve after the given time", assuming the test just started running
[12:03:28] Marking "should format seconds as a timestamp" as passed
[12:03:28] Marking "should return the seconds between 2 times" as passed
[12:03:28] Marking "should resolve after the given time" as passed
[12:03:28] No task result for "should match snapshot", assuming the test just started running
[12:03:28] Marking "should match snapshot" as passed
[12:03:28] Coverage was disabled due to all requests being exhausted
[12:03:28] Ending test run <none>
```

### Version

v1.2.10

### Validations

- [X] Check that you are using the latest version of the extension

BenShelton:
Fixes #477

It seems that when using workspace configs the coverage `enabled` flag does not get set to `true`, so the call to `initCoverageProvider` doesn't actually initialize the provider. This causes the report to not generate and coverage to not show.

There is also a similar issue within `waitForReport` where using the coverage config from the context directly doesn't reflect the correct `enabled` flag, so using the local one ensures the flag is set correctly.

I suppose the _correct_ fix here would be figuring out why this flag is not syncing correctly with workspaces, but this is a simple workaround.

This can be validated using the `monorepo-vitest-workspace` sample, after these changes the coverage UI in VSCode should show correctly as in screenshot:

Screenshot 2024-09-30 at 11 44 21

### DIFF
--- a/src/worker/coverage.ts
+++ b/src/worker/coverage.ts
@@ -16,7 +16,8 @@ export class VitestCoverage {
     private vitest: Vitest,
   ) {
     this._config = ctx.config.coverage
-    ctx.projects.forEach((project) => {
+    const projects = new Set([...ctx.projects, ctx.getCoreWorkspaceProject()])
+    projects.forEach((project) => {
       Object.defineProperty(project.config, 'coverage', {
         get: () => {
           return this.config
@@ -66,7 +67,6 @@ export class VitestCoverage {
     this.ctx.logger.log('Running coverage with configuration:', this.config)
 
     if (!this._provider) {
-      vitest.config.coverage.enabled = true
       // @ts-expect-error private method
       await vitest.initCoverageProvider()
       await vitest.coverageProvider?.clean(this._config.clean)
@@ -83,7 +83,7 @@ export class VitestCoverage {
   async waitForReport() {
     if (!this.enabled)
       return null
-    const coverage = this.config
+    const coverage = this.ctx.config.coverage
     if (!coverage.enabled || !this.ctx.coverageProvider)
       return null
     this.ctx.logger.error(`Waiting for the coverage report to generate: ${coverage.reportsDirectory}`)
--- a/src/worker/coverage.ts
+++ b/src/worker/coverage.ts
@@ -66,6 +66,7 @@ export class VitestCoverage {
     this.ctx.logger.log('Running coverage with configuration:', this.config)
 
     if (!this._provider) {
+      vitest.config.coverage.enabled = true
       // @ts-expect-error private method
       await vitest.initCoverageProvider()
       await vitest.coverageProvider?.clean(this._config.clean)
@@ -82,7 +83,7 @@ export class VitestCoverage {
   async waitForReport() {
     if (!this.enabled)
       return null
-    const coverage = this.ctx.config.coverage
+    const coverage = this.config
     if (!coverage.enabled || !this.ctx.coverageProvider)
       return null
     this.ctx.logger.error(`Waiting for the coverage report to generate: ${coverage.reportsDirectory}`)


src/worker/coverage.ts
```diff
@@ -66,6 +66,7 @@ export class VitestCoverage {
     this.ctx.logger.log('Running coverage with configuration:', this.config)
 
     if (!this._provider) {
+      vitest.config.coverage.enabled = true
```
sheremet-va, line 69: `enabled` is configured by `this.enabled`, it should be a getter referencing `this._enabled`

src/worker/coverage.ts
```diff
@@ -82,7 +83,7 @@ export class VitestCoverage {
   async waitForReport() {
     if (!this.enabled)
       return null
-    const coverage = this.ctx.config.coverage
+    const coverage = this.config
```
sheremet-va, line 86: `this.ctx.config.coverage` should already reference `this.config` (see the override at line 19)

sheremet-va (changes requested):

sheremet-va:
I think the problem here is that the `this.ctx.config` is not actually present in `this.projects` in your case and we need to reassign it manually. Something like this should work:

```ts
- ctx.prrojects.forEach(
+ ;[...ctx.projects, ctx.getCoreWorkspaceProject()].forEach(
```

We need to check if the core project is in projects first though

src/worker/coverage.ts
BenShelton: Agreed, will revert this.

BenShelton:
> I think the problem here is that the `this.ctx.config` is not actually present in `this.projects` in your case and we need to reassign it manually. Something like this should work:
> 
> ```diff
> - ctx.projects.forEach(
> + ;[...ctx.projects, ctx.getCoreWorkspaceProject()].forEach(
> ```
> 
> We need to check if the core project is in projects first though

You're right, that makes a lot of sense. Used a `Set` to avoid duplicating the projects and all seems to working correctly. Tested with the `vue` and `monorepo-vitest-workspace` samples and both are reporting coverage 😄

sheremet-va (approved):

sheremet-va:
Thank you!

Issue #515, opened 2024-10-29
Dynamically created test description not displayed in browser mode


kayahr:
### Describe the bug

If a test uses a dynamically created description (like `describe("sum" + " some context", ...`) then the testing panel  in VSCode shows the text `<unknown>` instead of the real text:

![image](https://github.com/user-attachments/assets/aba4d8e0-4782-4ca3-859c-053eb6e599d0)

But this only happens with enabled browser mode. When browser mode is disabled (running in node) then the description is shown directly. 

![image](https://github.com/user-attachments/assets/355d0d74-ff64-4abe-a8ea-b19692c22a9c)

When executing the tests then the `<unknown>` description is replaced with the correct one. But every time the tests are refreshed, the description vanishes again.

### Reproduction

* Clone https://github.com/kayahr/vitest-bugs-1
* Run `npm i`
* Open project in vscode with enabled vitest extension
* Open Testing panel.
* Click `Refresh tests` button and expand the tree. It shows text `<unknown>`
* Click `Run tests` button. Now the description is correct.
* Click `Refresh tests` again and the description is broken again. 

### Output

```shell
[INFO 5:02:38 PM] [v1.6.2] Vitest extension is activated because Vitest is installed or there is a Vite/Vitest config file in the workspace.
[INFO 5:02:38 PM] [API] Running Vitest v2.1.4 (vitest-bug-1/vitest.config.js) with Node.js@v22.10.0: /usr/bin/node 
[INFO 5:02:38 PM] [Worker] Port 63315 is in use, trying another one...
[INFO 5:02:38 PM] [Worker] Port 63316 is in use, trying another one...
[INFO 5:02:38 PM] [API] Vitest v2.1.4 (vitest-bug-1/vitest.config.js) child process 452285 created
[INFO 5:02:38 PM] [VSCODE] Watching vitest-bug-1 with pattern **/*
[INFO 5:02:38 PM] [API] Collecting tests: sum.test.js
[INFO 5:02:39 PM] [Worker] Collecting /home/k/Projects/vitest-bugs/vitest-bug-1/sum.test.js const __vite_ssr_import_0__ = await __vite_ssr_import__("/node_modules/vitest/dist/index.js", {"importedNames":["describe","it","expect"]});
const __vite_ssr_import_1__ = await __vite_ssr_import__("/sum.js", {"importedNames":["sum"]});



const someContext = "xyz";

__vite_ssr_import_0__.describe("sum " + "someContext", () => {
[...]

PR #518, opened 2024-10-30
fix: resolve the test message name statically when dynamic expression is used

sheremet-va:
Related #515

It is not possible to resolve the dynamic name during collection because it's done statically. but the name _will_ be resolved the next time the test has run.

This fixes the display, but running the test with a button won't work for these tests because we use `testNamePattern` to run tests.

We need to try and find a way to keep the original dynamic names when the collection is triggered. Maybe just ignore static names if the dynamic is known?

### DIFF
--- a/src/worker/collect.ts
+++ b/src/worker/collect.ts
@@ -1,6 +1,5 @@
 import type { SourceMap } from 'node:module'
 import { relative } from 'pathe'
-import type { TemplateLiteral } from 'acorn'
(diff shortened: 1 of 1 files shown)


kayahr:
Now that #519 is fixed (Thanks! Works perfectly now!) I no longer need my workaround with dynamic test names. So feel free to close this issue. If I understood it correctly it cannot be improved further anyway.

PR #545, opened 2024-11-27
Add stack traces to test message for test error.

ArthurHub:
- For thrown error scenarios provides the location of the error thrown inside the tested code with fully navigable trace.
- For non throw error scenarios provides additional context on the failed assertion.

|| Before | After |
|--------|--------|--------|
| Assert | assertion - before | assertion - after |
| Exception | exception - before | exception - after |


sheremet-va:
Looks good!

Issue #548, opened 2024-12-03
Debug test throws Error: No test suite found in file if project is located on D: drive


armbzk:
## Description

* Running `Debug test` in vscode vitest extension fails with `Error: No test suite found in file D:/gitea/armbzk/vitest-update/tests/dummy.test.mjs`
* Error only occurs with `Debug test` - `Run test` succeeds without any error
* after copying the project to Windows `C:` drive `Debug test` works fine
* `npx vitest run` in console succeeds (also on `D:` drive)

## Environment

* vscode 1.95.3
* OS: Windows_NT x64 10.0.26100 (Windows 11)
* Nodejs: 20.18.0
* npm 10.9.1
* vscode vitest extension v1.8.1
* vitest 2.1.7
* MSYS2 GNU bash, version 5.2.37(1)-release (x86_64-pc-msys)

### Reproduction

Link to gihub repo: https://github.com/armbzk/vitest-error

tests/dummy.test.mjs:

```js
import { describe, expect, it } from 'vitest';

describe('dummy tests', () => {
    it('should succeed', () => {
        expect(true).toBe(true);
    });
});
```

vitest.config.mts:

```ts
import { defineConfig } from 'vitest/config';

export default defineConfig({
    test: {
        environment: 'node',
        include: [
            'tests/*.test.{js,mjs}',
        ],
    },
});
```

package.json:

```json
{
  "devDependencies": {
    "vitest": "^2.1.7"
  },
  "scripts": {
    "test": "npx vitest run"
  }
}
```

### Output

```shell
Output in vscode DEBUG console:

C:\Program Files\nodejs\node.EXE c:\Users\armbzk\.vscode\extensions\vitest.explorer-1.8.1\dist\worker.js
Process exited with code 1

Error message in vitest test explorer:

Error: No test suite found in file D:/gitea/armbzk/vitest-update/tests/dummy.test.mjs
at runFiles (file:///digitea/armbzk/vitest-update/node_modules/@vitest/runner/dist/index.js:1254:11)
at startTests (file:///digitea/armbzk/vitest-update/node_modules/@vitest/runner/dist/index.js:1271:9)
at processTicksAndRejections (node:internal/process/task_queues:95:5)
at file:///digitea/armbzk/vitest-update/node_modules/vitest/dist/chunks/runBaseTests.3qpJUEJM.js:126:11
[...]

PR #577, opened 2025-01-21
fix: extract MemberExpression from SequenceExpression in getName

tsirlucas:


sheremet-va:
I don't think this is related to #548, but it does fix incorrect parsing when the browser mode is used, for example, or when the experimental option is set

Windows is failing, I will check it later this week. I think we can merge this in the meantime

tsirlucas:
>I don't think this is related to https://github.com/vitest-dev/vscode/issues/548
yeah after re-reading the error is similar but noticed everyone in that thread is related to windows and had issues even before vite 6. makes sense!

thank YOU for providing us such tool!

PR #603, opened 2025-03-21
feat(debug): support debugging with headed browser

nCastle1:
Fixes https://github.com/vitest-dev/vscode/discussions/474

This PR improves support for debugging in non-headless browser scenarios by adding the option to define an additional launch configuration to run after starting debugging. This enables connection to the browser running the tests, which enables hitting breakpoints.

Implementation details: after the debug session starts, if configured, a second debug session is launched as a child of the first. I confirmed by inspection of VS Code source that child debug sessions are stopped when the parent is stopped, so I don't think additional clean up is needed. ([Evidence](https://github.com/microsoft/vscode/blob/3286791a7e920c9b61b84fc2dbba136da912bb64/src/vs/workbench/contrib/debug/browser/debugSession.ts#L426))

I have included a sample project that demonstrates how to configure headful browser tests; it is copy-pasted version of the browser sample, with some config changes. I have also updated the readme to document this feature.

This is my first time contributing to this repo. I checked for contribution guidelines but didn't find anything beyond the code of conduct; please let me know if you have any other guidance.


sheremet-va:
I don't think we should require any new additional config options. We can infer them from the config ourselves.

nCastle1:
Thank you for that feedback, I will re-work the PR to infer as much as possible from the existing config.

src/debug.ts
```diff
@@ -183,6 +217,24 @@ async function getRuntimeOptions(pkg: VitestPackage) {
   }
 }
 
+function getAttachConfigForBrowser(config: ReturnType<typeof getConfig>, pkg: VitestPackage): vscode.DebugConfiguration {
```
nCastle1, line 220: I split this out originally with intent to support multiple browsers, but I'm not sure how feasible that is. Vitest only docs playwright with chromium for the vs code/IDE-integrated debugging. Firefox doesn't support inspect-brk flag, and the VS Code integration has some limitations that will make integration challenging - if the browser instance isn't running at attach time, the attach config fails immediately instead of waiting like the built-in chrome attach implementation.

src/worker/worker.ts
```diff
@@ -146,6 +146,10 @@ export class ExtensionWorker implements ExtensionWorkerTransport {
     return files.map(([project, spec]) => [project.config.name || '', spec])
   }
 
+  public getResolvedBrowserOptions(): ResolvedBrowserOptions | undefined {
+    return { ...this.ctx.config?.browser, commands: undefined }
```
nCastle1, line 150: This needs to be reworked, complex configuration (e.g. custom commands) will break the extension currently.

nCastle1:
I've updated the PR to take a zero configuration approach, reading the browser config as resolved by vitest. I have it working with playwright and chromium; this seems pretty solid. I'm working towards a solution for webdriverio, but I have zero experience with that technology so it is slow going.

I've noted a few key missing pieces, and I don't think this is in a mergable state, so I will keep as a draft. I will plan to revisit and add tests soon (pending some competing deadlines). Any feedback is very welcome in the meantime.

src/worker/worker.ts
sheremet-va: I think you can just rename this to `getBrowserDebugOptions` and return only needed fields

src/debug.ts
sheremet-va: I'd say if it doesn't work in Vitest CLI, the extension cannot add support for it, and waiting for it is a bit useless

It's better to throw an error, maybe? Or a warning?

nCastle1:
Thank you for the feedback so far and patience with the lingering PR.

I have updated the PR to work correctly in multi-project scenarios where some tests are browser mode and some are not. Debugging with playwright is working consistently (chromium only). Debugging with webdriverio is also sort-of working; I think pretty close to as well as it can work without enhancements to vitest.

I believe this PR is ready for another review.

samples/browser-webdriverio/test/.gitignore
```diff
@@ -0,0 +1 @@
+__snapshots__
```
sheremet-va, line 1: Can you also add screenshots to gitignore?

src/worker/worker.ts
```diff
@@ -146,6 +146,12 @@ export class ExtensionWorker implements ExtensionWorkerTransport {
     return files.map(([project, spec]) => [project.config.name || '', spec])
   }
 
+  public getBrowserDebugOptions(): (Pick<ResolvedBrowserOptions, 'enabled' | 'provider'> & { project: string })[] | undefined {
```
sheremet-va, line 149: I think it might be better to just return this in https://github.com/vitest-dev/vscode/blob/f0fa52bbb4acd425374b43972ebb8372431d8706/src/worker/index.ts#L51

Then you can store it in https://github.com/vitest-dev/vscode/blob/8fbf0ea83276472ca78eec464962ae0ec3152449/src/api.ts#L398

And get the values at any point

src/debug.ts
```diff
@@ -157,6 +214,32 @@ export async function debugTests(
...
+    return items
+  }
+  // Climb up tree until entry with project is found
+  if (item?.parent) {
+    return getProjectsFromTests(item.parent, api, tree)
+  }
+  return []
+}
+
+function getBrowserModeLaunchArgs(isPlaywright: boolean, config: any): string {
```
sheremet-va, line 235: Can we use an actual type instead of the `any`?

src/debug.ts
```diff
@@ -157,6 +214,32 @@ export async function debugTests(
...
 }
 
+function getTestProjectsInFolder(path: string, api: VitestFolderAPI, tree: TestTree) {
+  const folder = tree.getOrCreateFolderTestItem(api, path)
+  const items = tree.getFolderFiles(folder)
+  return items.map(item => (getTestData(item) as TestFile).project)
+}
+
+function getProjectsFromTests(item: vscode.TestItem | undefined, api: VitestFolderAPI, tree: TestTree): string[] {
+  const items = getTestProjectsInFolder(item?.uri?.fsPath ?? '', api, tree)
```
sheremet-va, line 224: It's quite hard to read a lot of `?.`, `??`, `| undefined` everywhere. Can this be simplified to only accept the value if it's there?

src/debug.ts
```diff
@@ -76,6 +79,13 @@ export async function debugTests(
     },
   }
 
+  // If the debug request includes any test files belonging the browser-mode projects,
+  // vitest needs to be started with the correct --inspect-brk and --browser arguments.
+  // Later, after debugging session starts, a secondary debug session is started; that session attaches to the launched browser instance.
+  const { browserModeProjects, isPlaywright } = await api.getBrowserModeInfo()
+  const testProjects = request.include?.filter(inc => inc.uri?.fsPath != null).flatMap(inc => getProjectsFromTests(inc, api, tree)) ?? []
+  const needsBrowserMode = !!browserModeProjects?.length && testProjects.some(project => browserModeProjects?.includes(project))
```
sheremet-va, line 87: Should we add a warning that non-chromium webdriverio tests are not supported?

src/debug.ts
```diff
@@ -157,6 +214,32 @@ export async function debugTests(
...
+  }
+  // Climb up tree until entry with project is found
+  if (item?.parent) {
+    return getProjectsFromTests(item.parent, api, tree)
+  }
+  return []
+}
+
+function getBrowserModeLaunchArgs(isPlaywright: boolean, config: any): string {
+  const browser = isPlaywright ? 'chromium' : 'chrome'
```
sheremet-va, line 236: I like that we provide the browser configuration ourselves. I think at one point there should be a UI in the extension to select the browser, but it's not related to these changes

src/debug.ts
```diff
@@ -157,6 +214,32 @@ export async function debugTests(
...
+  // Only playwright provider supports --inspect-brk currently
+  const inspectBrk = isPlaywright ? `--inspect-brk=localhost:${config.debuggerPort ?? '9229'}` : ''
+  // regardless of user config, some properties need to be set when debugging with browser mode enabled
+  return `vitest ${config.cliArguments ?? ''} ${inspectBrk} --browser=${browser}`
```
sheremet-va, line 240: What happens if `cliArguments` already have these options? They will be converted into arrays here:

https://github.com/vitest-dev/vscode/blob/f0fa52bbb4acd425374b43972ebb8372431d8706/src/worker/init.ts#L21

Maybe we should normalize them afterwards just in case?

sheremet-va (changes requested):
Looks very promising! Just a few suggestions. Thank you for your work on this 🙏🏻

src/debug.ts
```diff
@@ -182,6 +252,33 @@ export async function debugTests(
...
+  if (parentPath) {
+    return getProjectsFromTests(parentPath, parentItem?.parent, api, tree, excluded)
+  }
+  return []
+}
+
+function getBrowserModeLaunchArgs(isPlaywright: boolean, config: { debuggerPort?: number; cliArguments?: string }): string {
+  const browser = !config.cliArguments?.includes('--browser') ? `--browser=${isPlaywright ? 'chromium' : 'chrome'}` : ''
+  // Only playwright provider supports --inspect currently
+  const inspect = isPlaywright && !config.cliArguments?.includes('--inspect') ? `--inspect=localhost:${config.debuggerPort ?? '9229'}` : ''
```
nCastle1, line 277: Previously I was using `--inspect-brk`, but it seems that behavior is now consistent with `--inspect`. Now starting with `--inspect-brk` will pause on every file and not just the first, which is a bad experience.

I'm not really sure why this has changed; I reviewed release notes and recent PRs in Vite and Vitest but nothing jumped out.

nCastle1:
@sheremet-va sorry for the long delay in responding to your feedback, other projects had to take priority.

I've done my best to address all feedback. I've re-run tests locally and everything seems to be in order. Thank you in advance for your review.

src/debug.ts
```diff
@@ -165,15 +229,21 @@ export async function debugTests(
...
       disposables.reverse().forEach(d => d.dispose())
       disposables.length = 0
     }
   })
 
   const onDidTerminate = vscode.debug.onDidTerminateDebugSession((session) => {
-    if (session.configuration.__name !== 'Vitest')
+    // Child/secondary debug session should stop the main debugging session
+    if (session.parentSession?.configuration.__name === DebugSessionName) {
+      vscode.debug.stopDebugging(session.parentSession)
```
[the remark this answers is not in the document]
nCastle1: I've updated to check for the browser debug session specifically. I tested locally and it seems to work for both basic sample (it didn't work before for the reason you pointed out) and the browser samples.

src/debug.ts
```diff
@@ -86,6 +92,33 @@ export async function debugTests(
...
+  const excludedTestIds = new Set(request?.exclude?.map(ex => ex.id) ?? [])
+  const testProjects = includedTests?.filter(inc => inc.uri?.fsPath != null).flatMap(({ uri, parent }) => getProjectsFromTests(uri!.fsPath, parent, api, tree, excludedTestIds)) ?? []
+
+  const includedBrowserModeProjects = browserModeProjects?.filter(p => testProjects.includes(p.project))
+  const isPlaywright = !!includedBrowserModeProjects?.some(p => p.provider === 'playwright')
+  const isWebdriverio = !!includedBrowserModeProjects?.some(p => p.provider === 'webdriverio')
+
+  const needsBrowserMode = !!includedBrowserModeProjects?.length
+  if (needsBrowserMode) {
+    if (isWebdriverio && isPlaywright) {
```
nCastle1, line 111: I refined the messaging with more specific warnings, which seemed appropriate since the warnings will now be more prominent.

sheremet-va:
Clicking "Restart" seems to break the test runner

And I am getting this error after I try again:

Screenshot 2025-07-22 at 18 08 16

PR #616, opened 2025-05-21
chore: switch to tsdown

btea:


sheremet-va:
tsdown is more opinionated and needs extra configuration, it's not a simple replace, unfortunetly. For example, it uses `cjs` extension even if package has type `commonjs`. I will configure it myself in a separate PR

PR #650, opened 2025-08-12
feat: support Vitest 4

sheremet-va:
Vitest 4 removed previously deprecated APIs, so the extension also needs to be updated.


Aman007Agg:
I reviewed the task and marked it as **Fixable**.

Reasoning: the change is a substantial migration to support Vitest 4 alongside Vitest 3, but it remains within the scope of instructions/tests/oracle updates. The PR description and merged diff indicate a coordinated compatibility update rather than a fundamentally invalid task.

For the submitter form, I would answer:
- **What makes this task difficult?** The migration touches workspace structure, package boundaries, runtime worker selection, and backwards compatibility between Vitest 3 and 4 APIs, so solving it requires careful integration across multiple layers rather than a single code change.
- **Estimated senior-engineer time:** **40+ minutes**.

I did not mark it as invalid because the task appears technically feasible from the available metadata and diff context.

PR #679, opened 2025-11-03
chore: automatically build before launching debug sessions

felipecrs:
This avoids having to run `pnpm run dev` manually before starting debug sessions.


sheremet-va:
Not everyone has your configuration. This will break 99% of use cases.

PR #704, opened 2025-12-27
fix: Prevent a deadlock waiting for browser debugger to attach

Alanoll:
#702 fixed the fact that the debugger never attached to projects using the `project` style configuration, but I found in my larger repo's that a potential deadlock could occur waiting for the debugger to attach.

Scenario:
The RPC event `onBrowserDebug` is sent as soon as VSCode completes it's promise to attach to the browser in the [`then`](https://github.com/vitest-dev/vscode/blob/ff6e44d4841f59662c15ea67f69435edfb185e67/packages/extension/src/debug.ts#L202). There's a timing issue if the browser hasn't yet executed the `setupFile` from the extension that triggers the await for the event.

So:
1. Start Debug
2. Extension launches debug session for VSCode to attach to the browser
3. Vitest spins up the test project and initializes
4. VSCode completes attaching and triggers `onBrowserDebug` event
5. Browser loads the setupFile `browserSetupFile.ts` that await for the `onBrowserEvent` via the `__vscode_waitForDebugger` command created [here](https://github.com/vitest-dev/vscode/blob/ff6e44d4841f59662c15ea67f69435edfb185e67/packages/worker/src/reporter.ts#L50)

Because the event fired before the setupfile started to listen, the debug session deadlocks here waiting for the signal.

This PR starts listening to the event before adding the `setupFile` to the project collection, and stores the state in a private variable. The Command now checks this variable first to see if it's already attached (and immediately returns) or subscribes to the event again like it was doing before to resolve the promise.

### DIFF
--- a/packages/worker/src/reporter.ts
+++ b/packages/worker/src/reporter.ts
@@ -25,7 +25,7 @@ export class VSCodeReporter implements Reporter {
   private setupFilePaths: WorkerInitMetadata['setupFilePaths']
   private debug: WorkerRunnerOptions['debug']
 
-  private debuggerAttached: boolean = false
+  private debuggerAttached: boolean | undefined = undefined
 
   constructor(options: VSCodeReporterOptions) {
     this.setupFilePaths = options.setupFilePaths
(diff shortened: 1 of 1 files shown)


Alanoll:
@sheremet-va This (if approved) I think might need to be in the same release as #702

sheremet-va (changes requested):

PR #707, opened 2026-01-06
feat: use reporters in the test results panel

GermanJablo:
### Problem
Prior to this pull request, the test results panel only displayed the `console.logs` that occurred during the test.
Clicking on the test instance in the right panel displays the expected and actual values ​​in separate boxes, along with a series of traces. However, this format is divided into different terminals, making it difficult to copy and paste, or even to read (for both humans and AI).

### Solution
With this PR, two things are displayed:
1. The first line shows the command to replicate in the CLI.
2. Then, the complete test output, as defined in the user configuration's `reporters`.

### Before:

image

### After:

image

#### 1. `runner.ts` - Preserve ANSI colors & clean output
- **Remove `stripVTControlCharacters()`**: VS Code's Test Results panel supports ANSI escape codes for colors
- **Delay TestRun cleanup**: Prevents premature closure of the output stream while async results are still arriving
- **Clean output**: Automatically strips Vitest's watch mode prompts ("Watching for file changes...") to keep the log focused on results
- **Show CLI command**: Displays the equivalent command for easy reproducibility from the terminal
- **Skip duplicate console.logs**: Avoids double-printing logs that reporters already handles

#### 2. `rpc.ts` - Add persistent `onProcessLog` handler
- **Create and preserve handler**: Output from worker arrives via RPC. If the handler is cleared between runs, subsequent executions have nowhere to send stdout/stderr

#### 3. `child_process.ts` - Connect RPC to stdout callbacks
- **Forward `onProcessLog` to callbacks**: The worker sends output via WebSocket/RPC, not Node's native stdout. This bridges RPC events to the stdout callback system

#### 4. `worker/index.ts` - Enable DefaultReporter
- **Add 'default' reporter**: Run Vitest's `DefaultReporter` alongside `VSCodeReporter` to get CLI-style output
- **Respect user reporters**: If user has custom reporters configured (e.g., `reporters: ['verbose']`), those are used instead of 'default'

### Minor Formatting Differences

Due to the way Vitest's detects terminal environments, there are small formatting differences compared to CLI output:

- **Extension (without TTY)**: No suite hierarchy (e.g., `❯ addition (9)` not shown)
- **columns/separators**: the number of columns available in the terminal, used for example in separators, is not calculated.

### DIFF
--- a/packages/worker/src/index.ts
+++ b/packages/worker/src/index.ts
@@ -112,7 +112,6 @@ export async function initVitest(
 
             return {
               test: {
-                printConsoleTrace: true,
                 coverage: {
                   reportOnFailure: true,
                   reportsDirectory: join(tmpdir(), `vitest-coverage-${randomUUID()}`),


packages/extension/src/api/child_process.ts
```diff
@@ -95,11 +95,14 @@ export async function createVitestProcess(pkg: VitestPackage) {
         resolved.handlers.onStdout = (callback: (data: string) => void) => {
           stdoutCallbacks.add(callback)
         }
-        const clearListeners = resolved.handlers.clearListeners
-        resolved.handlers.clearListeners = () => {
-          clearListeners()
-          stdoutCallbacks.clear()
-        }
+
+        // Forward RPC process logs to the same stdout callbacks
```
sheremet-va: why?

packages/extension/src/runner.ts
```diff
@@ -160,37 +160,16 @@ export class TestRunner extends vscode.Disposable {
...
-        log.info('[TEST]', consoleLog.content)
+      // Signal that the test run is complete, but DON'T set this.testRun to undefined yet
+      // The testRun will be properly ended in the finally block of startTestRun
+      if (!collecting) {
+        this.testRunDefer?.resolve()
       }
     })
 
+    // Skip individual console logs since DefaultReporter already includes them
+    // api.onConsoleLog((consoleLog) => { ... }
```
sheremet-va: What is this?

packages/worker/src/index.ts
```diff
@@ -136,10 +145,6 @@ export async function initVitest(
         },
       ],
     },
-    {
-      stderr,
-      stdout,
-    },
   )
```
sheremet-va: Why

sheremet-va (commented):
These changes seems to break things rather than fix them. Please open an issue with the desired functionality, it needs to be discussed first

packages/extension/src/runner.ts
```diff
@@ -797,7 +793,7 @@ function formatTestPattern(tests: readonly vscode.TestItem[]) {
 }
 
 function formatTestOutput(output: string) {
-  return stripVTControlCharacters(output.replace(/(?<!\r)\n/g, '\r\n'))
+  return output.replace(/(?<!\r)\n/g, '\r\n')
```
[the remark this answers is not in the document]
sheremet-va: Colors are not supported in inline view, at least

GermanJablo:
> These changes seems to break things rather than fix them. Please open an issue with the desired functionality, it needs to be discussed first

Thanks for the thorough review! You're right, I rushed the PR, and there might be things I'm not addressing correctly.

I've opened [this issue](https://github.com/vitest-dev/vscode/issues/709) hoping we can agree on the "why" before moving on to the "how."

Thank you so much for your time!

PR #742, opened 2026-03-03
Fix: change `@vitest/browser/context` import to `vitest/browser`

nstepien:
This should resolve the following warning when debugging tests:
> DEPRECATED  C:/Users/me/.vscode/extensions/vitest.explorer-1.44.1/dist/browserSetupFile.mjs tries to load a deprecated "@vitest/browser/context" module. This import will stop working in the next major version. Please, use "vitest/browser" instead.

### DIFF
--- a/packages/extension/src/worker/browserSetupFile.ts
+++ b/packages/extension/src/worker/browserSetupFile.ts
@@ -1,4 +1,4 @@
-import { commands, server } from '@vitest/browser/context'
+import { commands, server } from 'vitest/browser'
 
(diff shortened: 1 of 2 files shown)


nstepien:
I see. We can keep this PR open until Vitest 3 support gets removed 👍

sheremet-va:
> I see. We can keep this PR open until Vitest 3 support gets removed 👍

This won't happen for a while, the goal of the extension is to support all versions starting with 1.4.0

PR #797, opened 2026-07-11
feat: set diagnostic.source="Vitest"

BroadlyWhitaker:
Set the source field to "Vitest" when inserting new diagnostics into vscode.


sheremet-va (approved):
Thanks!

PR #803, opened 2026-08-06
fix: run gutter tests with the most specific config when several configs claim the same file

### LINKED ISSUE #799: Vitest extension doesn't run test files with tags
### Describe the bug

I am using [tags](https://vitest.dev/guide/test-tags) in my unit tests and noticed that the [Vitest extension for Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=vitest.explorer) cannot run tests that are having tags declared.

**Screenshot:**

Image

### Expected behavior

Tests are being executed.

### Reproduction

Build a test case with tags, i.e.

```ts
describe('getResultOrThrow', () => {
  it('identifies an up-trending asset by a positive ROC', {tags: ['tulipindicators']}, () => {
    // ...  
  });
});
```

The extension cannot run the test although the tags are defined in the `vites.config.ts` file:

```ts
import { defineConfig, mergeConfig } from "vitest/config";
import baseConfig from "../../vitest.config.base";

export default mergeConfig(
  baseConfig,
  defineConfig({
    test: {
      tags: [
        {
          description: "...",
          name: "tulipindicators",
        },
      ],
    },
  }),
);
```
 
My `vitest.config.base.ts` file looks like this:

```ts
import {defineConfig} from 'vitest/config';

export default defineConfig({
  esbuild: {
    /*
     * Allows using the "accessor" keyword in TypeScript:
     * https://github.com/vitest-dev/vitest/issues/5976#issuecomment-2190804966
     */
    target: 'es2022',
  },
  test: {
    bail: 1,
    coverage: {
      exclude: ['src/start/**'],
      include: ['src/**'],
      provider: 'v8',
      reporter: ['html', 'lcov', 'text'],
    },
    environment: 'node',
    globals: true,
    // https://main.vitest.dev/guide/learn/writing-tests-with-ai.html#mock-cleanup
    restoreMocks: true,
    sequence: {
      concurrent: true,
      shuffle: true,
    },
  },
});
```

### Output

```shell
> Error: The Vitest config does't define any "tags", cannot apply "tulipindicators" tag for this test. See: https://vitest.dev/guide/test-tags
```

### Extension Version

1.50.8

### Vitest Version

4.1.10

### Validations

- [x] Check that you are using the latest version of the extension
- [x] Check that there isn't [already an issue](https://github.com/vitest-dev/vscode/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitest-dev/vscode/discussions) or join our [Discord Chat Server](https://chat.vitest.dev).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

bennycode:
Fixes #799

A shared base config like `vitest.config.base.ts` matches the extension's `configGlob`, so it gets its own Vitest instance that claims the same test files as the package-level configs that merge it. The gutter "Run Test" button then runs tests with the base config, which is missing the merged options (`tags`, `env`, …):

```
Error: The Vitest config does't define any "tags", cannot apply "tulipindicators" tag for this test.
```

## Why registration order alone doesn't fix it

@sheremet-va suggested in #799 to register the profiles in a different order so vscode picks up the more specific config automatically. I tried that first and on its own it changes nothing. A file item only gets the tag of the first config that resolves it (`getOrCreateFileTestItem` returns the cached item without touching its tags), and since profiles are tag-scoped, the shallowest config's profile is the only one VS Code can ever pick for that file, no matter the registration order.

## Fix

1. File items collect the tags of every config that claims them. `getOrCreateFolderTestItem` already does this for folder items. File items were the inconsistency: with two configs claiming the same directory, the folder item is runnable by both configs' profiles, but the files inside it only by the first one.
2. Run/debug/coverage profiles are registered for all discovered configs before any process spawns, deepest config first. When several profiles can run the same item, vscode invokes the first registered one, so the most specific config becomes the effective default. Label and tag come from the config path, so nothing waits for resolution and initialisation isn't delayed (the concern raised in #799). Config resolution order and workspace config claiming are unchanged; profiles for configs that never resolve are disposed.

The base config's profile stays available in the profile picker; only the effective default changes. Configs at the same depth behave as before.

## Tests

Added `samples/shared-base-config` and an e2e test that clicks the gutter run button. The sample test only passes under the package-level config (it asserts on an `env` value), so the run result shows which config executed it: `0/1` before this change, `1/1` after.


bennycode:
Closing because the proposed change only makes all claiming profiles eligible; it does not change which profile VS Code selects by default and therefore does not resolve #799.

PR #811, opened 2026-09-03
feat: support opening trace view

hi-ogawa:
- part of https://github.com/vitest-dev/vitest/issues/10769
- requires https://github.com/vitest-dev/vitest/pull/11190
  - not strictly required since non-focused layout is still usable.

This PR add per-test actions to open trace view focused layout of html report in an extension-managed webview. The webview tab is reused when switching tests and auto reselects/refreshes when the report is regenerated.

There are some design choices, but for now, the mechanism is tried to be kept simple. Future follow-ups may include:
- add to command-palette to open the trace view based on last run (not from the explicitly selected test)
- use artifact data in extension host directly to build trace view UI instead of relying on html reporter output.
- use watch mode UI (or use `onTestArtifactRecord`) so it can renders streamed trace view of running test.
- show trace line location marker in editor gutters

_Example_

https://github.com/user-attachments/assets/d5edb829-f9ee-47be-943c-2a2c26a79b3e


packages/extension/src/api.ts
```diff
@@ -192,8 +192,11 @@ async function createVitestProcessAPI(
         usedConfigs.add(project.config)
       }
     })
-    const files = await meta.rpc.getFiles()
-    const config = new VitestProjectConfig(pkg, meta.projects, meta.workspaceSource)
+    const [files, workerConfig] = await Promise.all([
+      meta.rpc.getFiles(),
+      meta.rpc.getExtensionConfig(),
```
sheremet-va: You should already have config data by this point. Pass it down when the worker is initialized, like we pass down other config options

sheremet-va (changes requested):

packages/extension/src/api.ts
hi-ogawa: That's what I've keep nudging agents to find. You don't believe initial mess 🤣  https://github.com/vitest-dev/vscode/pull/811/changes/dcb65a13f9288e45a0083504f2c18f5c9ac9abb7 Will keep nudging.

hi-ogawa: It looks like passing through new value from vitest to extension takes some hops. I've refactored it first so adding one more html reporter thing becomes cleaner.
- https://github.com/vitest-dev/vscode/pull/815

pnpm-workspace.yaml
```diff
@@ -56,6 +56,10 @@ catalogs:
     '@vitest/utils': ^3.2.4
     vite: ^7.2.6
     vitest: ^3.2.4
+  v5:
+    '@vitest/browser-playwright': 5.0.0-rc.4
+    '@vitest/ui': 5.0.0-rc.4
+    vitest: 5.0.0-rc.4
```
hi-ogawa: For now pinning `5.0.0-rc.4` because v5 stable seems to have issues with extensions.

hi-ogawa: Well, looks like `5.0.0` works fine now. It looks like the issue was maybe some other dependencies of `samples/browser-v5/package.json`.

hi-ogawa (commented):
Actually there might be a big ergonomic issues with multiple test cases. This requires users to re-opening browser view for each test. 🤔

sheremet-va:
> Actually there might be a big ergonomic issues with multiple test cases. This requires users to re-opening browser view for each test. 🤔

Ideally we should have a webview like playwright has for the best DX. I had a prototype for it, but never pushed

hi-ogawa:
> > Actually there might be a big ergonomic issues with multiple test cases. This requires users to re-opening browser view for each test. 🤔
> 
> Ideally we should have a webview like playwright has for the best DX. I had a prototype for it, but never pushed

Push that please 😃

sheremet-va:
> > > Actually there might be a big ergonomic issues with multiple test cases. This requires users to re-opening browser view for each test. 🤔
> > 
> > 
> > Ideally we should have a webview like playwright has for the best DX. I had a prototype for it, but never pushed
> 
> Push that please 😃

Had to merge conflicts, not sure if it works: https://github.com/vitest-dev/vscode/pull/818

The initial idea was to have it for settings, but we can enhance it. Feel free to use the code from there, I don't think I will finish it 😄

packages/worker/package.json
```diff
@@ -7,6 +7,7 @@
   },
   "devDependencies": {
     "@vitest/utils": "catalog:latest",
+    "vite": "catalog:latest",
```
hi-ogawa, line 10: Not sure why, but this was needed for typecheck. Otherwise it looks like type check loads Vite 7 and Vite 8 at the same time somewhere.

README.md
```diff
@@ -134,6 +135,28 @@ If you hover over it, you can get a more detailed diagnostic.
...
+### Trace View
+
+With Vitest 5, the extension can open recorded browser interactions for a selected test in [Vitest Trace View](https://vitest.dev/guide/browser/trace-view.html). Enable Trace View and the HTML reporter in your existing Browser Mode configuration:
+
+```ts
+import { configDefaults, defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  test: {
+    reporters: [...configDefaults.reporters, 'html'],
```
sheremet-va, line 147: In general I don't like asking the user to update their configs since we can set any config we like when initiating Vitest, but I think it's fine for the initial version.

sheremet-va (approved):
Looks dope!

PR #815, opened 2026-09-04
refactor: group worker config metadata

hi-ogawa:
- preliminary cleanup for https://github.com/vitest-dev/vscode/pull/811

This defines a bag of config-related data that is passed from Vitest worker to extension host as `ExtensionWorkerConfig`. This would make it simpler for https://github.com/vitest-dev/vscode/pull/811 to introduce one more value for html reporter path in the same bag.

### DIFF
--- a/packages/extension/src/apiProcess.ts
+++ b/packages/extension/src/apiProcess.ts
@@ -1,4 +1,4 @@
-import type { SerializedProject } from 'vitest-vscode-shared'
+import type { ExtensionWorkerConfig, SerializedProject } from 'vitest-vscode-shared'
 import type { VitestPackage } from './spawn/pkg'
 import { usesJestTestNamePattern } from './spawn/pkg'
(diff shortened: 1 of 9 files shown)


packages/extension/src/apiProcess.ts
```diff
@@ -22,8 +22,7 @@ export class VitestProjectConfig {
 
   constructor(
     readonly pkg: VitestPackage,
-    readonly projects: SerializedProject[],
-    readonly workspaceSource: string | false,
+    readonly extensionConfig: ExtensionWorkerConfig,
```
sheremet-va: This doesn’t seem like the right name. Extension config is something that configures the extension

hi-ogawa: Naming is hard. What I ended up is `metadata: WorkerReadyMetadata`. Now we have to live with `meta.metadata` in some places.

sheremet-va (approved):

PR #820, opened 2026-09-08
feat: add Bun runtime support

N0SAFE:
## Description

Add support for running Vitest tests with the Bun runtime. Bun v1.4 added native Vitest support including `--coverage` with threads and forks pools ([blog post](https://bun.com/blog/bun-v1.4#vitest)).

Closes #473

## Changes

- **Configuration**: Added `'bun'` to `vitest.runtime` enum (`'auto' | 'node' | 'deno' | 'bun'`)
- **Type Safety**: Extended `WorkerInitMetadata.runtime` and `VitestPackage.runtime` types to include `'bun'`
- **Auto-detection**: Updated `guessRuntime()` to auto-detect Bun projects by checking for `bun.lock`, `bun.lockb`, or `bunfig.toml`
- **Process Spawning**: Added `--bun` flag when spawning Bun processes (this flag tells Bun to replace Node.js APIs with native implementations)
- **Runtime Resolution**: Extended `findRuntimeExecutable()` to resolve the `bun` executable
- **Debug Support**: Added `--bun` runtime argument for debugging with Bun
- **Version Guard**: Added minimum Bun version check (>= 1.4.0) - Bun versions below 1.4 will show a warning and skip the package
- **Sample Project**: Added a Bun sample project at `samples/bun/` for testing

## How it works

When `vitest.runtime` is set to `'bun'` (or auto-detected), the extension will:
1. Find the `bun` executable on the system PATH
2. Validate that Bun version >= 1.4.0 (required for Vitest support)
3. Pass the `--bun` flag when spawning the Vitest worker process
4. Use JSON serialization for RPC communication (since Bun's `v8.serialize` uses JavaScriptCore's wire format, not V8's format)

The `--bun` flag is critical — it makes Bun replace Node.js APIs with its native implementations, which is required for Vitest compatibility.

## v8.serialize/v8.deserialize Note

Bun does have `v8.serialize` and `v8.deserialize` functions, but they use JavaScriptCore's wire format instead of V8's binary format. This means data serialized with Bun is NOT compatible with Node.js, so we use JSON serialization (same approach as Deno).

## Testing

- Type checking passes (`pnpm typecheck`)
- Build succeeds (`pnpm build`)
- Unit tests pass (17/17)

## References

- [Discussion #473](https://github.com/vitest-dev/vscode/discussions/473) - Original feature request
- [Bun v1.4 Blog Post](https://bun.com/blog/bun-v1.4#vitest) - Bun's native Vitest support announcement
- [oven-sh/bun#4145](https://github.com/oven-sh/bun/issues/4145) - Original Bun Vitest support issue (now resolved)

### DIFF
--- a/packages/extension/src/utils.ts
+++ b/packages/extension/src/utils.ts
@@ -4,6 +4,7 @@ import { spawn } from 'node:child_process'
 import fs from 'node:fs'
 import { inspect, stripVTControlCharacters } from 'node:util'
 import { dirname, relative } from 'pathe'
+import { gte } from 'semver'
 import * as vscode from 'vscode'
 import which from 'which'
 import { getConfig } from './config'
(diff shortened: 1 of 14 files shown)


packages/extension/src/utils.ts
```diff
@@ -32,6 +33,47 @@ export function pluralize(count: number, singular: string) {
   return `${count} ${singular}${count === 1 ? '' : 's'}`
 }
 
+export async function getBunVersion(cwd: string): Promise<string | null> {
```
sheremet-va, line 36: I don't think we need to check bun version. We don't check node or deno

sheremet-va (changes requested):

packages/extension/src/utils.ts
N0SAFE: I understand, but Bun needs to be at least version 1.4. Otherwise, there are APIs that were previously unsupported, the Vitest code coverage run will fail unless Bun is upgraded to version 1.4 or later.

